PR #1, opened 2017-06-05
Pull #1: simplest generator

Luolc:
I created a maven project.

The `config` folder, `.travis.yml`, `LICENSE`, `pom.xml`, `.gitignore` refer to checkstyle/checkstyle and checkstyle/contribution repo.

Current structure:
- git: Contains git related classes.
  - DiffParser: Git diff parser. Parse the git diff between a specific branch and master, and return a list of GitChange.
  - GitChange: Represents git changes of a file. Currently it only has path info. There would be more details in the future.
- generator: Contains regression test config generators and utility classes.
  - AbstractConfigGenerator: The base class for config generators. Contains the common initialization and generation logic.
  - ConfigGeneratorFactory: Defines a factory API that enables developers to obtain a generator that processes the git changes and generate config XML.
  - DefaultConfigGenerator: The simplest generator implementation using only default options of checks.

`AbstractConfigGenerator` has two important methods: `createModuleElements` and `whetherToSkipChange`. The implementations of generator would override these two.

`DefaultConfigGenerator` is one of the implementation. As we early discussed in my proposal, it is only concerned about the changes of checkstyle module files and generate config with default options.

### DIFF
--- a/config/checkstyle.properties
+++ b/config/checkstyle.properties
@@ -1,0 +1,4 @@
+checkstyle.suppressions.file=config/suppressions.xml
+checkstyle.header.file=https://raw.githubusercontent.com/checkstyle/checkstyle/master/config/java.header
+checkstyle.regexp.header.file=https://raw.githubusercontent.com/checkstyle/checkstyle/master/config/java_regexp.header
+checkstyle.importcontrol.file=config/import-control.xml
--- a/src/main/java/com/github/checkstyle/regression/generator/DefaultConfigGenerator.java
+++ b/src/main/java/com/github/checkstyle/regression/generator/DefaultConfigGenerator.java
@@ -1,0 +1,91 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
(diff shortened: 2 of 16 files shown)


config/checkstyle.properties
```diff
@@ -0,0 +1,4 @@
+checkstyle.suppressions.file=config/suppressions.xml
+checkstyle.header.file=https://raw.githubusercontent.com/checkstyle/checkstyle/master/config/java.header
+checkstyle.regexp.header.file=https://raw.githubusercontent.com/checkstyle/checkstyle/master/config/java_regexp.header
+checkstyle.importcontrol.file=config/import-control.xml
```
rnveach: Please add new lines to all files that don't end with one if it doesn't cause any issues.
I see atleast 4 files missing them.

rnveach:
@Luolc We just enabled travis for this repo, so next push should start validation as long as it is setup correctly.
To make reviews easier and faster, lets split this PR into 2.
1st PR will just be empty src folders but have everything else for maven project and to test it with travis. I will merge this quickly.
2nd PR will be all new java code you have added.

Missing code coverage is fine for now, but we will very quickly enforce that.

Please also create issues for your work, just like we do in main project.
It will help me follow what you are planning to do next.
For 2nd PR, please create issue for it. 1st PR is fine without issue.

rnveach:
Also, leave empty `Main` in 1st PR. Checkstyle will probably complain if it doesn't have a file to check.

rnveach:
@Luolc As I am looking over the java code in a bit more detail, please split them apart again between `git` and `generator` and their issues and PRs.
Each area should be pretty isolated and almost mini-programs that just pass data between each of them.

src/main/java/com/github/checkstyle/regression/generator/DefaultConfigGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return returnValue;
+    }
+
+    /**
+     * Creates the class corresponding to the changed file path.
+     * @param path the changed file path
+     * @return the class corresponding to the changed file path
+     */
+    private Class<?> createClassFromPath(String path) {
```
rnveach: When you create issue for this, we will need to re-think this method.
We are already importing checkstyle release dependency in maven.
How are we going to load PR's snapshot checkstyle than? How can we be sure we are loading snapshot class from memory and not release?

We can't have 2 versions of a class in memory at the same time.
We also need to compare Checkstyle PR snapshot to previous Checkstyle snapshot.

rnveach:
other PR was merged.
Let's close this and start fresh with new PR with `git` component.

Issue #8, opened 2017-06-06
Add more to CI


rnveach:
We need to add code coverage. This is a priority.

We should split checkstyle and unit tests and code coverage to their own travis items so we get parallel running.
We should add sevntu checks.
We should add findbugs and pmd.

We need to add commit validation test from main repo.

rnveach:
All items are done and merged.

PR #9, opened 2017-06-06
Issue #8: added cobertura code coverage

rnveach:
Issue #8

Added code coverage verification.
Main and project is set to 0 for now. Coverage should increase with every new code to main, I hope.
It doesn't have to be 100%.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,4 +81,79 @@
             </plugin>
         </plugins>
     </build>
+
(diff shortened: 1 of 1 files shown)


rnveach:
https://travis-ci.org/checkstyle/regression-tool/builds/239921067#L609-L612
Cobertura ran and passed.

Issue #3, opened 2017-06-05
Git component


rnveach:
Taken from https://github.com/checkstyle/regression-tool/pull/1 ,

We need to analyze `git` repository for changes in new commits in PR branch.

Notes:
* We most ignore deleted files. Deleted checks and such can never be checked for changes as they are gone. Moved/Renamed files should stay to make sure there is no issue using them.
* We must only grab files from `src/main`. Any changes in tests/it should be ignored.

Should we also grab the contents of the changes? Will they be needed?

Restricting files further by type should be done elsewhere as we will have to look at class hierarchy and such.

rnveach:
@Luolc Based on discussions in PR, we will not be dropping `src/main` or deleted files. We will try to pull them in for #5 .
Will Git component do anything else?

Luolc:
@rnveach 
I agree with your comment at https://github.com/checkstyle/regression-tool/pull/17#pullrequestreview-43132180 that we would expends the fields in the future, including file path and changed lines.
From https://github.com/checkstyle/regression-tool/pull/17#issuecomment-307365616, 
> If you wish to separate this out more, than filtered files would be a parameter to this class.

Do you mean add some regex strings to this class? Would the filters only depend on path? I am not very sure.

The mainly missing thing is the changed lines in GitChange, and maybe we could add them when we implement a generator which need them.

rnveach:
> mean add some regex strings to this class?

That or full paths, but lets wait on this until we see a specific need.
I can see filtering on path and change type. I still think deleted files won't give us any benfits, but we'll wait and see.

> maybe we could add them when we implement a generator which need them.

That sounds fine to me.

If you see nothing else to add to the core of the git component, than I think we can close this issue.

Luolc:
@rnveach  I think it is OK to close the issue now.

romani:
just a note:
please do not be crazy with changes detection, it is better to grow its ability to detect changes with ability to process it.

rnveach:
> it is better to grow its ability to detect

That is how we are breaking things apart.

Issue #2, opened 2017-06-05
Discuss components involved in regression


rnveach:
Discuss and agree on all components project should have at a high level.

By PR, we currently have Git and Generator.

rnveach:
I agree with git component.

Should we have `checkstyle` component?
We need to identify the type of file to restrict to only modules (for now). In the future, we also need to identify if the module has new/removed properties as they will need to be included in the regression.

I think we should rename `generator` to `configuration` as it will be clearer what we are doing in the package, creating the configuration(s).

We will need `report` component, which runs the configuration(s) and pushes the reports to a server.

IMO, each component should be completely isolated. We should have an extra package `data` that will be used to pass data between all the components.

Luolc:
For `module` component. I agree with the functions you listed. But as pointed at https://github.com/checkstyle/regression-tool/pull/1#pullrequestreview-42136513. The method was not a good implementation, for it just instantiate the class for the release version of checkstyle, and that is meaningless. I haven't think out a way to load the class from file and meanwhile avoid conflicts by now. Should we discuss it here or create new issue? 

> I think we should rename generator to configuration as it will be clearer what we are doing in the package

I am OK with it.

> We will need report component, which runs the configuration(s) and pushes the reports to a server.

OK. Should we reuse the existing code in contribution repo?

rnveach:
> Should we discuss it here or create new issue?

Create new issue, this discussion is for only high level information. Specifics will be done in specific issue for each component.
If we agree on high level details for each component, than create a new issue for all of them except the `data` package.
Put in all your specifics, questions, issues, and we will discuss things further.

> Should we reuse the existing code in contribution repo?

Be more specific, I assume you mean the regression groovy scripts? If so, than yes, we should try to re-use them.
We need a good reason not to be able to use them as they are widely used now.

Luolc:
I am thinking about the filter. In git component there would be filters depending on path and changed type. But we may need some other filter, i.e. filter to judge whether it is a checkstyle modole. So we might need filters in `configuration` module as well. What's your option?

rnveach:
IMO, filtering based on module should be in `module` component. `git` shouldn't be trying to dedude checkstyle modules as that puts too much work into the one module.
`configuration` should be given only the information it needs to create the configuration(s). So all the filtering should be done by then.

Luolc:
@rnveach 
[...]

PR #17, opened 2017-06-08
Issue #3: Git component

Luolc:
#3

### DIFF
--- a/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
+++ b/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
@@ -1,0 +1,110 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.git;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.diff.DiffEntry;
+import org.eclipse.jgit.lib.Constants;
+import org.eclipse.jgit.lib.ObjectReader;
+import org.eclipse.jgit.lib.Ref;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.revwalk.RevTree;
+import org.eclipse.jgit.revwalk.RevWalk;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+import org.eclipse.jgit.treewalk.AbstractTreeIterator;
+import org.eclipse.jgit.treewalk.CanonicalTreeParser;
+
+/**
+ * Parses git diff between feature branch and master for the further use.
+ * @author LuoLiangchen
+ */
+public final class DiffParser {
+    /** Prevents instantiation. */
+    private DiffParser() {
+    }
+
+    /**
+     * Parses the diff between a given branch and the master in the give repository path.
+     * @param repositoryPath the path of checkstyle repository
+     * @param branchName     the name of the branch to be compared with master
+     * @return a list of {@link GitChange} to represent the changes
+     * @throws IOException     JGit library exception
+     * @throws GitAPIException JGit library exception
+     */
+    public static List<GitChange> parse(String repositoryPath, String branchName)
+            throws IOException, GitAPIException {
+        final File gitDir = new File(repositoryPath, ".git");
+        final Repository repository = new FileRepositoryBuilder().setGitDir(gitDir)
+                .readEnvironment().findGitDir().build();
+        final Git git = new Git(repository);
+        final AbstractTreeIterator featureTreeParser =
+                prepareTreeParser(repository, Constants.R_HEADS + branchName);
+        final AbstractTreeIterator masterTreeParser =
+                prepareTreeParser(repository, Constants.R_HEADS + "master");
+        final List<GitChange> returnValue = git.diff()
+                .setOldTree(masterTreeParser)
+                .setNewTree(featureTreeParser)
+                .call()
+                .stream()
+                .map(DiffParser::convertDiffEntryToGitChange)
+                .collect(Collectors.toList());
+        git.close();
+        repository.close();
+        return returnValue;
+    }
+
+    /**
+     * Creates a tree parser from a commit, to be used by diff command.
+     * @param repository the repository to parse diff
+     * @param ref        the name of the ref to the commit; e.g., "refs/heads/master"
+     * @return the tree parser
+     * @throws IOException JGit library exception
+     */
+    private static AbstractTreeIterator prepareTreeParser(Repository repository, String ref)
+            throws IOException {
+        final Ref head = repository.exactRef(ref);
+        final RevWalk walk = new RevWalk(repository);
+        final RevCommit commit = walk.parseCommit(head.getObjectId());
+        final RevTree tree = walk.parseTree(commit.getTree().getId());
+        final CanonicalTreeParser treeParser = new CanonicalTreeParser();
+        final ObjectReader reader = repository.newObjectReader();
+        treeParser.reset(reader, tree.getId());
+        reader.close();
+        walk.dispose();
+        walk.close();
+        return treeParser;
+    }
+
+    /**
+     * Converts a {@link DiffEntry} to {@link GitChange} for the further use.
+     * @param diffEntry the {@link DiffEntry} instance to be converted
+     * @return the {@link GitChange} instance converted from the given {@link DiffEntry}
+     */
+    private static GitChange convertDiffEntryToGitChange(DiffEntry diffEntry) {
+        return new GitChange(diffEntry.getNewPath());
+    }
+}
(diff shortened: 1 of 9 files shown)


src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -0,0 +1,106 @@
...
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.revwalk.RevTree;
+import org.eclipse.jgit.revwalk.RevWalk;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+import org.eclipse.jgit.treewalk.AbstractTreeIterator;
+import org.eclipse.jgit.treewalk.CanonicalTreeParser;
+
+/**
+ * Git diff parser.
```
rnveach: Repeating the component and class name is not ok for a description.
Just write a brief sentence that states the purpose of the utility.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
```diff
@@ -0,0 +1,81 @@
...
+            if (!file.createNewFile()) {
+                throw new IOException("Could not create file " + file);
+            }
+            git.add().addFilepattern(fileName).call();
+            git.commit().setMessage("add " + fileName).call();
+            return file;
+        }
+    }
+
+    public static void addAllAndCommit(Repository repository, String message)
```
rnveach: why `all`? that usually implies more than 1 of something.
I only see one file being added and it is named `.`.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
```diff
@@ -0,0 +1,81 @@
...
+        final Repository repository = FileRepositoryBuilder.create(new File(repoDir, ".git"));
+        repository.create();
+        return repository;
+    }
+
+    public static void createNewBranchAndCheckout(Repository repository, String branchName)
+            throws GitAPIException {
+        try (Git git = new Git(repository)) {
+            if (git.branchList().call().stream()
+                    .anyMatch(ref -> ref.getName().equals("refs/heads/" + branchName))) {
```
rnveach: Use the constants here too.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -0,0 +1,60 @@
...
+        GitUtils.addAllAndCommit(repository, "append text to HelloWorld");
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        repository.close();
+    }
+
+    @Test
+    public void testParse() throws Exception {
```
rnveach: We should include all types of differences in git.
New file, removed file, renamed file, moved file, chmod change (linux only).

rnveach: We should also add a test where PR branch is behind master. This happens quite a bit when people don't stay up to date with changes in master.
We should still correctly identify PR branch commits in this scenario and only get it's differences.

rnveach: We also need a test where PR branch has multiple commits on master.

Luolc:
@rnveach 
Sorry for the missing of my comments. It was pretty late when I sent this PR yesterday (about 2AM) and I was sort of sleepy then. I originally intended to edit this PR to add my opinions today.

I had thought about the issue comments before this PR and decided to hang on "deleted files" and "src/main" for the following reasons:

Regrading "src/main", I have discussed the possibility of grabing information from changes of UTs in my proposal with Roman. I summarized and wrote the result of the discussion in the proposal as:
> Moreover, from changes of UTs we could take values of properties that are required for regression. As certain UT is changed, it means that behavior for that property values are changed and to do regression on values that UT specify is required. Therefore, we could find the changed lines of corresponding UTs, and then find the config setting code, which could be used to help us generate the configurations. It might be not easy, but I think it is possible and we could have a try.

https://docs.google.com/document/d/1xu6SE4qeKTRQ45R9FSLOQB-t5ExzBGyvU9FLGifvxY0/edit?usp=sharing, Project -> Outline -> 2 

This idea is much far from now, and even may not be achieved during the period of GSoC. But there are chances to realize the idea in the future. So I think we should not drop "it/test" and other files here in git component, but in specific strategy class. 

In my original design of `Generator`, there is a `whetherToSkipChange` method for the changed files filter. And I prefer smth like that to do the filtering outside git component. Or we could create an filter interface in git component.
https://github.com/Luolc/regression-tool/blob/74c6c83094599488bb9960d5d7eb162858a3753d/src/main/java/com/github/checkstyle/regression/generator/AbstractConfigGenerator.java#L183-L190

In addition, from the point of [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle), I think there should be only git related operation in git component. DiffParser, then only parse the diff, no other restriction. If we add "it/test" filter, then git component surely dependent on the file structure of checkstyle project. At least it would increase the difficulty of testing. 

The situation of "deleted files" is similar.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -0,0 +1,106 @@
...
+ * @author LuoLiangchen
+ */
+public class DiffParser {
+    /** The prefix of branch head commit ref. */
+    private static final String COMMIT_REF_PREFIX = "refs/heads/";
```

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
Luolc: Done.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
Luolc: Here it is the same with command `git add .`, and that means add all files.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
Luolc: Updated.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Contains utility methods for git component test.
+ * @author LuoLiangchen
+ */
+public final class GitUtils {
+    /** Prevents instantiation. */
+    private GitUtils() {
+    }
+
+    public static Repository newRepository() throws IOException {
```
Luolc: Done.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Parses the diff between a given branch and the master.
+     * @param branchName the name of the branch to be compared with master
+     * @return a list of {@code GitChange} to represent the changes
+     * @throws IOException JGit library exception
+     * @throws GitAPIException JGit library exception
+     */
+    public List<GitChange> parse(String branchName) throws IOException, GitAPIException {
```
Luolc: I just designed it by my intuition previously. But actually my point is on the opposite, I think if we need work with more than one repository, then static method is better. 
I was thinking that we may work with only one repository path and more than one branch, for there is only one checkstyle project but many feature branches. Then we could just set the repo path once in the constructor and run `parse` method with only the branch parameter.
But I guess in most case we would work with only one repo path and one branch. So if you think static method is better, I am ok with that.

Luolc:
@rnveach 
Regrading the test, I suggest to split the remaining testing problems to separate issues, for it is a little hard for me to complete all of them in this one PR. :) And let me fix them one by one in the future.
BTW, I don't know why the "static method" review is folded. I replied my points in the review block.

rnveach:
> then it surely dependents on the file structure of checkstyle project

Whole regression idea depends on checkstyle project.
If you wish to separate this out more, than filtered files would be a parameter to this class.

> "src/main"

Ok, we will hold off on this for now.
Changes in tests can take a priority, but since we know the module that is changing, we can deduce the class' test and pull all properties from the file. If we rely only on test changes and for whatever reason there is no test changes found, we still need to come up with some type of testing plan.

This will be done in the module component then?

>  suggest to split the remaining testing problems to separate issues

Make the issue. They can all be in one issue and worked on slowly since they are all very closely related. We have no requirement to fix whole issue in this repo. Just keep issue updated with what is left.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
rnveach: IMO, I think the 2 branches we will have to work with is PR branch and master branch. Since we can't install the Checkstyle versions, we will need to execute them to get the information we need (what are modules, what are properties, and what has changed).

But in terms of GitDifferencing, we only need the files that have changed (for now).
Let's make this static and remove the field and make the constructor private. I don't forsee any components having a constructor, right now.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
rnveach: I was unaware, as I don't use the git command line.
This can be ignored for now.

src/main/java/com/github/checkstyle/regression/git/GitChange.java
```diff
@@ -0,0 +1,47 @@
...
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.git;
+
+import org.eclipse.jgit.diff.DiffEntry;
```
rnveach: This may be another issue we can do once I see how this class will be used.

------

If we are passing this class to other components, than `org.eclipse.jgit` can't be in this class and this class must go to `com.github.checkstyle.regression.data` as I defined in #2 .

I don't want jgit showing up in import controls of other components as it makes it look like they are doing things outside of their purview.
I think we should expand the fields to what we need, the file path and the lines that have changed.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Parses the diff between a given branch and the master.
+     * @param branchName the name of the branch to be compared with master
+     * @return a list of {@link GitChange} to represent the changes
+     * @throws IOException JGit library exception
+     * @throws GitAPIException JGit library exception
+     */
+    public List<GitChange> parse(String branchName) throws IOException, GitAPIException {
```
rnveach: Based on discussion at https://github.com/checkstyle/regression-tool/pull/17#discussion_r120983450 ,

> Let's make this static and remove the field and make the constructor private. I don't forsee any components having a constructor, right now.

src/main/java/com/github/checkstyle/regression/git/GitChange.java
Luolc: I remove the DiffEntry from the constructor params. The path and changed lines (in the future) would be passed directly now.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java

Issue #18, opened 2017-06-11
Refine DiffParserTest


Luolc:
Taken from https://github.com/checkstyle/regression-tool/pull/17#pullrequestreview-42994141

- We should include all types of differences in git.
New file, removed file, renamed file, moved file, chmod change (linux only).
- We should also add a test where PR branch is behind master. This happens quite a bit when people don't stay up to date with changes in master.
We should still correctly identify PR branch commits in this scenario and only get it's differences.
- We also need a test where PR branch has multiple commits on master.

rnveach:
This should be done before or right after Issue #15 as we want this utility very well tested before we give it to Checkstyle for actual usage.
Tests will show if our code works correctly for the strange cases we will find in real usage.

rnveach:
Message from Luolc (connectivity issues):
> For `PR branch is behind master`:
> I did some investigation on issue 18 today. When master is latest and PR doesn't rebase latest code, there would be diff that we don't expect. Git CLI has the same behavior, so we have to find a workaround to handle not latest PR. I don't know if we could do this without rebase currently. I will keep investigating.

Luolc:
The git CLI has the same behavior. So it is a feature of git to show the diff of un-rebased files. While in GitHub "File Changed" tab, unrelated files won't appear. So it is sure to have a way to avoid this. At least we could run rebase command for it, but it should be the last solution.

Luolc:
I find a command `git diff master...<branch>` can fit our need, but I don't know how to implement this in JGit yet.

Here is an example to see how it works:
```bash
$ git init
$ 
$ echo HelloWorld > HelloWorld
$ git add .
$ git commit -m "initial commit"
$ 
$ git checkout -b foo
$ 
$ git checkout master
$ echo ChangeInMaster > ChangeInMaster
$ git add .
$ git commit -m "change in master"
$ 
$ git checkout foo
$ echo ChangeInFoo > changeInFoo
$ git add .
$ git commit -m "change in foo"
$ 
$ git diff master
diff --git a/ChangeInFoo b/ChangeInFoo
new file mode 100644
index 0000000..7e38b29
--- /dev/null
+++ b/ChangeInFoo
@@ -0,0 +1 @@
+ChangeInFoo
diff --git a/ChangeInMaster b/ChangeInMaster
deleted file mode 100644
index f284658..0000000
--- a/ChangeInMaster
+++ /dev/null
@@ -1 +0,0 @@
-ChangeInMaster
$
[...]

Issue #19, opened 2017-06-11
Module Component 


Luolc:
Taken from #2 
We need to identify the type of file to restrict to only modules (for now). In the future, we also need to identify if the module has new/removed properties as they will need to be included in the regression.

Luolc:
As pointed at https://github.com/checkstyle/regression-tool/pull/1#pullrequestreview-42136513. The method was not a good implementation, for it just instantiate the class in the release version of checkstyle, and that is meaningless. I haven't think out a way to load the class from file and meanwhile avoid conflicts by now.

Luolc:
For the branch version of checkstyle, we only have the java source code. If we want to use the `ModuleReflectionUtils.isCheckstyleModule` way, then we must install the branch version. So we have to do some initialization(compile and install the branch version) before we run the regression-tool. Even in this way we are not able to load checkstyle stable version and branch version in the meantime.
A more interesting question is how to deal with the change of classes like `ModuleReflectionUtils`, as they are used in the regression-tool project. It seems a infinite cycle, well, I don't know how to describe it in English.

From other point of view, I think we have some other way to avoid loading the branch version file by now. Currently, our goal is to judge whether a class is a checkstyle module. And we have the hardcode map in `PackageObjectFactory`. We could grab this info from the content of `PackageObejctFactory`. Maybe a better way is to split the map into a separate xml, then we could load the xml directly.

Luolc:
According to https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/com/puppycrawl/tools/checkstyle/configuration_1_3.dtd, we have `module`, `property` and `message` element. (I don't know what's the use of `metadata`).
1. Is it possible to generate config with `message` element? I haven't used that in regression by now, so I am not clear.
2. What is the use of `default` key of property element? I don't see it in the config of checkstyle repo.
```
<!ELEMENT property EMPTY>
<!ATTLIST property
    name NMTOKEN #REQUIRED
    value CDATA #REQUIRED
    default CDATA #IMPLIED
>
```

rnveach:
I advise against asking free form questions in issue as it will cloud main discussion. This sounds like it would be better in gitter or google groups.

> I don't know what's the use of metadata

See https://github.com/checkstyle/checkstyle/issues/2726 .

> Is it possible to generate config with message element?

Configuration can override module's message.

> What is the use of default key of property element? 

[...]

Issue #15, opened 2017-06-06
NoException Checkstyle Project Regression


rnveach:
We need to consider running the snapshot version of the jar against the real Checkstyle project as a final test to make sure any new changes work without throwing an exception.

The question is what commit should we run against? Only the latest one or the top X? Should we create our own commit to test some basic avenues?

Reports shouldn't be uploaded, regardless.

This is completely separate from the junits.

This is blocked by default module testing.

These regression(s) will also showcase to Checkstyle project how to use our utility and incorporate it into their process.

rnveach:
@Luolc Please come up with a way to attach our tool to checkstyle in travis' CI.
There is no documentation, so this is the only way others will know how to use our utility without viewing the source.

Luolc:
@rnveach Got it. I would create a issue for documentation and write it at meantime.

Luolc:
@rnveach 
I think I may need to enable Travis-CI on my fork of checkstyle project and test this on it. What's your opinion? This could avoid bothering the dev on checkstyle project, but maybe not convenient for reviewing.

rnveach:
@Luolc Feel free to create your own travis. I do this too in checkstyle.

We still need this in our CI here before we incorporate it in any external place like checkstyle.

rnveach:
Fix was merged.

PR #22, opened 2017-06-12
Issue #18: Refine DiffParserTest

Luolc:
#18

- [x] Changed Type ADD, MODIFY, DELETE, RENAME, COPY
- [x] chomd change
- [x] PR branch behind master
- [x] PR branch has multiple commits on master

IMO, PR branch behind master is not necessary. If the PR is not able to merge (has conflict), then we have no need to run the regression-tool at all. If the PR is able to merge, then we could do `git rebase master` first to let the PR branch ahead of master.

### DIFF
--- a/src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
+++ b/src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
@@ -23,13 +23,16 @@ import static com.github.checkstyle.regression.internal.TestUtils.assertUtilsCla
 import static org.junit.Assert.assertEquals;
 
 import java.io.File;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.PosixFilePermission;
 import java.util.List;
+import java.util.Set;
 
+import org.apache.commons.lang.SystemUtils;
 import org.eclipse.jgit.lib.Repository;
 import org.junit.After;
-import org.junit.Before;
 import org.junit.Test;
 
 import com.github.checkstyle.regression.data.GitChange;
@@ -37,20 +40,8 @@ import com.github.checkstyle.regression.data.ImmutableGitChange;
 import com.github.checkstyle.regression.internal.GitUtils;
 
 public class DiffParserTest {
-    private Repository repository;
-
-    @Before
-    public void setUp() throws Exception {
-        repository = GitUtils.createNewRepository();
-        final File helloWorld = GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
-        GitUtils.createNewBranchAndCheckout(repository, "foo");
-        Files.write(helloWorld.toPath(), "hello world!".getBytes(), StandardOpenOption.APPEND);
-        GitUtils.addAllAndCommit(repository, "append text to HelloWorld");
-    }
-
     @After
     public void tearDown() throws Exception {
-        repository.close();
         GitUtils.clearTempRepositories();
     }
 
@@ -60,11 +51,37 @@ public class DiffParserTest {
     }
 
     @Test
-    public void testParse() throws Exception {
-        final List<GitChange> changes = DiffParser.parse(
-                repository.getDirectory().getParent(), "foo");
-        assertEquals(1, changes.size());
-        assertEquals("HelloWorld", changes.iterator().next().path());
+    public void testParseAddChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.addAnEmptyFileAndCommit(repository, "AddedFile");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be 1 change detected", 1, changes.size());
+            final GitChange expected = ImmutableGitChange.builder()
+                    .path("AddedFile")
+                    .build();
+            assertEquals("The change is not as expected", expected, changes.get(0));
+        }
+    }
+
+    @Test
+    public void testParseModifyChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            final File helloWorld = GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            Files.write(helloWorld.toPath(), "hello world!".getBytes(Charset.forName("UTF-8")),
+                    StandardOpenOption.APPEND);
+            GitUtils.addAllAndCommit(repository, "append text to HelloWorld");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be 1 change detected", 1, changes.size());
+            final GitChange expected = ImmutableGitChange.builder()
+                    .path("HelloWorld")
+                    .build();
+            assertEquals("The change is not as expected", expected, changes.get(0));
+        }
     }
 
     @Test
@@ -77,6 +94,50 @@ public class DiffParserTest {
             final List<GitChange> changes = DiffParser.parse(
                     repository.getDirectory().getParent(), "foo");
             assertEquals("There should be no changes detected", 0, changes.size());
+        }
+    }
+
+    @Test
+    public void testParseRenameChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorldFoo");
+            GitUtils.removeFileAndCommit(repository, "HelloWorld");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be 1 change detected", 1, changes.size());
+            final GitChange expected = ImmutableGitChange.builder()
+                    .path("HelloWorldFoo")
+                    .build();
+            assertEquals("The change is not as expected", expected, changes.get(0));
+        }
+    }
+
+    @Test
+    public void testParseCopyChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "a.txt");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.addAnEmptyFileAndCommit(repository, "b.txt");
+            GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/c.java");
+            GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/d.java");
+            GitUtils.removeFileAndCommit(repository, "a.txt");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be 3 change detected", 3, changes.size());
+            final GitChange expected0 = ImmutableGitChange.builder()
+                    .path("b.txt")
+                    .build();
+            assertEquals("The change is not as expected", expected0, changes.get(0));
+            final GitChange expected1 = ImmutableGitChange.builder()
+                    .path("src/com/foo/c.java")
+                    .build();
+            assertEquals("The change is not as expected", expected1, changes.get(1));
+            final GitChange expected2 = ImmutableGitChange.builder()
+                    .path("src/com/foo/d.java")
+                    .build();
+            assertEquals("The change is not as expected", expected2, changes.get(2));
         }
     }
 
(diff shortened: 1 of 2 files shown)


rnveach:
> is not necessary

We should still have this test to verify correct behavior will happen. As I said before, this is a common occurrence.

> If the PR is able to merge, then we could do git rebase master first

You will still need a test to verify this will work correctly as the rebase action should still be part of the tool. We can't assume the repository will be given to us in a perfect state.

> git rebase

I am slightly against the rebase idea as this will require modifying the branch and users may want to run our tool locally before a PR is even created to see regression themselves. We may also need local running to debug any issues. (This made me think of something else for a new issue.)
It will also prevent us from running this tool on old commits to see regression behavior, making Issue #15 harder to test.

The idea of this test was just to verify we get the correct differences of just the new commits and not any subtractions from the newer commits in master.
If we can achieve this without doing an actual rebase, than I would prefer that way, otherwise we can open an issue and discuss the rebase more.

Luolc:
@rnveach Got it. I would investigate and think over the "ahead master" situation and try to find a way to achieve this with rebase. I assume current code would lead to subtractions. :(

rnveach:
> I assume current code would lead to subtractions.

We can't assume anything, but yes, this is why I specifically want this test in.
If test proves this, than make it into a separate issue.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
```diff
@@ -63,6 +67,7 @@ public static File addAnEmptyFileAndCommit(Repository repository, String fileNam
             throws IOException, GitAPIException {
         try (Git git = new Git(repository)) {
             final File file = new File(repository.getDirectory().getParent(), fileName);
+            file.getParentFile().mkdirs();
```
rnveach: Please explain this.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
```diff
@@ -75,8 +80,23 @@ public static File addAnEmptyFileAndCommit(Repository repository, String fileNam
     public static void addAllAndCommit(Repository repository, String message)
             throws GitAPIException {
         try (Git git = new Git(repository)) {
+            final Status status = git.status().call();
```
rnveach: Please explain this.

Luolc: In git CLI, `git add .` add all kinds of change, including deleted. But in JGit, the `add .` could not add the deleted files into staging area. To obtain the same behavior as git CLI, we have to use `RmCommand` to handle deleted files.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
Luolc: I wrote a `GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/c.java")`, and it would fail if there is no `src/com/foo` directory existing. I am not very sure if this implementation is appropriate, for I am not very familiar with Java NIO in the past.

src/main/java/com/github/checkstyle/regression/git/GitChange.java
```diff
@@ -19,6 +19,8 @@
 
 package com.github.checkstyle.regression.git;
 
+import org.eclipse.jgit.diff.DiffEntry;
```
Luolc: Sorry I forgot it again. :( 
So your suggestion is move `GitChange` to `data` package? I am OK with it. Or maybe one more option: create a `git.data` package and move it there. Just come up another option and I have no preference on these two. :) I would wait your opinion and do the refactor then.

There is a question that whether we should expand a `ChangeType` field in `GitChange` in production code. I think that depends on whether we would treat them the same way or different way. Expect for `DELETE` (the file is gone), should we just not care about it is a new file or a modified file? By now the default config generator does not care about it.

If we decide to not introduce `ChangeType` to `GitChange` now. Then we could split `parse` method into two: one process the repository and return a list of JGit `DiffEntry`; the other convert `DiffEntry` to `GitChange`. Then we could test the `ChangeType` by calling first method.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
rnveach: Add a comment before the new code stating this, **but first**,

I commented out this code and got no test failures on Windows. I can't check Linux right now.
Please confirm you get a failure without it.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -20,47 +20,112 @@
...
+    public void testParseDeleteChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.addAnEmptyFileAndCommit(repository, "FileToDelete");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.removeFileAndCommit(repository, "FileToDelete");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals(1, changes.size());
+            assertChange(DELETE, "/dev/null", changes.get(0));
```
rnveach: Where did `/dev/null` come from?
The file that was deleted was `FileToDelete` so I expect that to be the file name.
This file name makes no sense if you are planning to parse/use the file name for anything.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
rnveach: I see now, we are passing a string to create a file in a specific path.
This is ok.

src/main/java/com/github/checkstyle/regression/git/GitChange.java
rnveach: > So your suggestion is move GitChange to data package?

Yes, but I don't expect us to do this in a test only issue like I also stated before.

-------

For your other questions, they shouldn't be asked/answered in a test only PR/issue.
We will lose this discussion later on.
I will expand on this in a new post below.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
Luolc: https://github.com/eclipse/jgit/blob/005e5feb4ecd08c4e4d141a38b9e7942accb3212/org.eclipse.jgit/src/org/eclipse/jgit/diff/DiffEntry.java#L373

In our previous discussion I want to keep "src/test" and deleted files mainly because I thought it is better to keep the possibility of getting all the changes, and leave a filter interface to let the real strategy code do decide which type of changes to use.

But I am re-thinking to drop deleted files now. If we would treat "ADD" and "MODIFY" the same way, then there is no need to distinguish the ChangeType. And "DELETE" is for sure useless as the file is gone.

rnveach:
@Luolc All your questions on deleted files are valid, but we haven't even made concrete plans yet in main issues where the question of their use would come into play, like at https://github.com/checkstyle/regression-tool/issues/19 . We **MUST** cement ideas on logic in these issues before we can say how code should change. We can't just change code without thinking things through a bit more.

----

For this PR, this was only to add more tests and not changing production code so much.
Let's focus on making tests here and not changing production code. If we see wrong behavior in production code, let's create a new issue and fix them there.

Changes in DiffParser for renames might need to be looked at more. If a Module is renamed, do we need to know its before name along with it's new name?
Let's wait on changes in `GitChange` until we see a need for knowing the type of change in production code. I can see us maybe needing to know if it is a rename or delete. I can't see us caring about the others.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
Luolc: There are no test failures for we don't have deleting operation in UTs (I mean no `File.delete` directly, but only remove files with `RmCommand`). 
I add this because I find this information on stack overflow, or somewhere else anyway, and I test it and find it correct. It doesn't affect our UTs by now, but the real behavior is not fitting the method name `addAllAndCommit`. The behavior is smth like `addAllExcludingDeletedFilesAndCommit`. And for most case we want the former, like what we usually do in git CLI(command `git add .`)

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
rnveach: Make a new issue for tracking deleted changes and give it no assignment. Mention this test.
We will discuss it there and possibly later on. We need to see what information we can get from these changes.

rnveach: > https://github.com/eclipse/jgit/blob/005e5feb4ecd08c4e4d141a38b9e7942accb3212/org.eclipse.jgit/src/org/eclipse/jgit/diff/DiffEntry.java#L373

Yes, you have to use `getOldPath`. So if we want to keep deleted files, this file name is a bug.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
rnveach: Ok. Please add the comment before it. I am fine with this.

Luolc: Comment added.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
Luolc: issue #25 for this

rnveach:
Please rebase over latest master and lets remove changes to production and make issues for incorrect behavior found in tests.

Luolc:
@rnveach I would rebase once #23 is done.

rnveach:
> splitting parse method into two to appease test of ChangeType, if we don't want to introduce ChangeType into GitChange class now.

We have no use for change type, so I don't see a reason to add it.
Tests will just show file names which is what every other component will be seeing.

rnveach:
#23 has been merged.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -23,45 +23,106 @@
...
+    @Test
+    public void testParseRenameChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorldFoo");
+            GitUtils.removeFileAndCommit(repository, "HelloWorld");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals(2, changes.size());
```
Luolc: Should we detect RENAME change? In current production code we would only have ADD, MODIFY and DELETE. And "one RENAME" would be seen as "one DELETE + one ADD". If we turn on "detectRename", then there would be only one change with type RENAME.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -23,45 +23,106 @@
...
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "a.txt");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.addAnEmptyFileAndCommit(repository, "b.txt");
+            GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/c.java");
+            GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/d.java");
+            GitUtils.removeFileAndCommit(repository, "a.txt");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals(4, changes.size());
```
Luolc: Same as above.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
rnveach: > And "one RENAME" would be seen as "one DELETE + one ADD". 

Based on discussion at https://github.com/checkstyle/regression-tool/issues/25 , we will be ignoring `DELETE`.
Will we be processing a `RENAME` differently than a `ADD`? Right now, I think we should always run full regression regardless.

So we can ignore renames for now.

rnveach:
@Luolc ping

rnveach:
@Luolc Is this ready for review? You never mentioned anything that I can see.

Luolc:
@rnveach Not yet. "PR branch behind master" is split to issue #45 and sent as PR #46 .

rnveach:
@Luolc If it is a separate issue and PR, why can't we finish this one? It has been open a long time.
Nothing listed here had to do with PR behind master.

Luolc:
@rnveach OK I would work on this at the meantime.

Luolc:
"PR branch behind master" is done at #45 
"chmod" change and PR with multiple commits are done.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -98,4 +159,48 @@ public void testParsePrBranchBehindMaster() throws Exception {
...
+                final File file = GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+                GitUtils.createNewBranchAndCheckout(repository, "foo");
+                final Set<PosixFilePermission> permissions =
+                        Files.getPosixFilePermissions(file.toPath());
+                permissions.add(PosixFilePermission.OWNER_EXECUTE);
+                Files.setPosixFilePermissions(file.toPath(), permissions);
+                GitUtils.addAllAndCommit(repository, "Change file mode");
+                final List<GitChange> changes = DiffParser.parse(
+                        repository.getDirectory().getParent(), "foo");
+                assertEquals("There should be 1 change detected", 1, changes.size());
```
rnveach: If the only change is a chmod (no changes to the contents of the file), I think we should ignore the change altogether.
Chmod has no bearing on the final contents of the JAR, which doesn't support chmod.
If you agree, make a new issue for it. We will require a new test in that issue, chmod plus file change in 1 commit.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -81,6 +98,50 @@ public void testParseDeleteChange() throws Exception {
...
+
+    @Test
+    public void testParseCopyChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "a.txt");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.addAnEmptyFileAndCommit(repository, "b.txt");
+            GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/c.java");
+            GitUtils.addAnEmptyFileAndCommit(repository, "src/com/foo/d.java");
+            GitUtils.removeFileAndCommit(repository, "a.txt");
```
rnveach: None of these files have contents, so how can we tell which one is being copied where?
Also `a` is removed, so this is suppose to act like another rename? If so, the test only says `Copy`.
If we are copying `a` to `b` and `c`, what is purpose of `d`?

Luolc: I was puzzled when I first investigating on `COPY`. It has the following rules:
- If no files deleted, then anything new would be detected as `ADD`.
- If one file is deleted, then one file added, and they are the "same"(in Git context), then it is a `RENAME`.
- If one file is deleted, some files added, and they the "same", then one of them is a `RENAME`, others is `COPY`.

> how can we tell which one is being copied where?

I don't exactly know, since the "same" in Git is describing as `score`. I guess the `RENAME` is one with the highest score and been added to staging area earliest.

Actually this UT is almost a copy of https://github.com/eclipse/jgit/blob/005e5feb4ecd08c4e4d141a38b9e7942accb3212/org.eclipse.jgit.test/tst/org/eclipse/jgit/diff/RenameDetectorTest.java#L209

rnveach: When I debug `DiffParser` everything in `DiffEntry` is recognized as `ADD` for me.
Is this just because we don't turn on recognizing copies?
When I open the repsository in EGit, and squash the commits, it does say 
copy' and 'rename'.
If I do remove, `Delete a.txt` then EGit says there is no `copies`.
It is weird behavior.

Let's ignore this for now.

Luolc: Yep, I removed the code for detecting `RENAME` and `COPY`. Now all of them would be detected as `ADD`.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
Luolc: @rnveach #49 for this.

Issue #24, opened 2017-06-13
GitChange should be moved to data package


rnveach:
https://github.com/checkstyle/regression-tool/blob/4edeee4de35edfeb48bdc030ea1d0d0d93e834ab/src/main/java/com/github/checkstyle/regression/git/GitChange.java

Data passed between components should be located in the `data` package.
They should be simple POJOs, with `final` fields, and try to avoid special classes like jgit.
It should be rare to update the import control for this package.

Luolc:
Taken from the discussion in https://github.com/checkstyle/regression-tool/pull/22#pullrequestreview-43498710:
There is a question that whether we should expand a `ChangeType` field in `GitChange` in production code. Not regrading test code, I think that depends on whether we would treat the changes the same way or different way in config generators. Expect for `DELETE` (the file is gone), should we just not care about whether it is a new file or a modified file? By now the default config generator does not care about it.

rnveach:
> Should we only focus on moving GitChange to data package without other changes in this issue?

Yes, this issue should only be us moving the class and not adding more fields.
We still have no need for the type of change in any other production code right now.

rnveach:
Change was merged.

PR #26, opened 2017-06-13
Issue #24: GitChange should be moved to data package

Luolc:
#24


config/import-control.xml
```diff
@@ -8,5 +8,6 @@
     <allow pkg="java.util"/>
     <subpackage name="git">
         <allow pkg="org.eclipse.jgit"/>
+        <allow pkg="com.github.checkstyle.regression.data"/>
```
rnveach: all components will be using this package, so lets move it to root.

Issue #23, opened 2017-06-12
DiffParser.parse should close/free resources on exceptions


rnveach:
https://github.com/checkstyle/regression-tool/blob/4edeee4de35edfeb48bdc030ea1d0d0d93e834ab/src/main/java/com/github/checkstyle/regression/git/DiffParser.java#L58-L78
https://github.com/checkstyle/regression-tool/blob/4edeee4de35edfeb48bdc030ea1d0d0d93e834ab/src/main/java/com/github/checkstyle/regression/git/DiffParser.java#L87-L100

We should always close resources, even if an exception occurs, to free up any locked resources.
Code should be changed to `try-finally` (not `try-with-resource`).
Powermock can be used to make exceptions happen and appease code coverage if required.

rnveach:
Fix is merged

PR #27, opened 2017-06-14
Issue #23: DiffParser.parse should close/free resources on exceptions

Luolc:
#23

### DIFF
--- a/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
+++ b/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
@@ -59,23 +59,35 @@ public final class DiffParser {
      */
     public static List<GitChange> parse(String repositoryPath, String branchName)
             throws IOException, GitAPIException {
+        final List<GitChange> returnValue;
         final File gitDir = new File(repositoryPath, ".git");
         final Repository repository = new FileRepositoryBuilder().setGitDir(gitDir)
                 .readEnvironment().findGitDir().build();
-        final Git git = new Git(repository);
-        final AbstractTreeIterator featureTreeParser =
-                prepareTreeParser(repository, Constants.R_HEADS + branchName);
-        final AbstractTreeIterator masterTreeParser =
-                prepareTreeParser(repository, Constants.R_HEADS + "master");
-        final List<GitChange> returnValue = git.diff()
-                .setOldTree(masterTreeParser)
-                .setNewTree(featureTreeParser)
-                .call()
-                .stream()
-                .map(DiffParser::convertDiffEntryToGitChange)
-                .collect(Collectors.toList());
-        git.close();
-        repository.close();
+
+        try {
+            final Git git = new Git(repository);
+
+            try {
+                final AbstractTreeIterator featureTreeParser =
+                        prepareTreeParser(repository, Constants.R_HEADS + branchName);
+                final AbstractTreeIterator masterTreeParser =
+                        prepareTreeParser(repository, Constants.R_HEADS + "master");
+                returnValue = git.diff()
+                        .setOldTree(masterTreeParser)
+                        .setNewTree(featureTreeParser)
+                        .call()
+                        .stream()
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -57,25 +57,36 @@ private DiffParser() {
      * @throws IOException     JGit library exception
      * @throws GitAPIException JGit library exception
      */
+    @SuppressWarnings("TryFinallyCanBeTryWithResources")
```
rnveach: I'm not sure what utility is generating this, but lets just turn it off instead of adding suppress warnings every time.
We will most likely never do this in this project as it is too hard to support code coverage.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -86,19 +97,31 @@ private DiffParser() {
...
-        treeParser.reset(reader, tree.getId());
-        reader.close();
-        walk.dispose();
-        walk.close();
-        return treeParser;
+        try {
+            final Ref head = repository.exactRef(ref);
+            final RevCommit commit = walk.parseCommit(head.getObjectId());
+            final RevTree tree = walk.parseTree(commit.getTree().getId());
+            returnValue = new CanonicalTreeParser();
```
rnveach: even this doesn't throw an exception, I feel it should still be inside the try.
If that causes trouble, than attach it to `returnValue`'s declaration, or at the very least move it down one line closer to try.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -86,19 +97,31 @@ private DiffParser() {
...
-        final RevCommit commit = walk.parseCommit(head.getObjectId());
-        final RevTree tree = walk.parseTree(commit.getTree().getId());
-        final CanonicalTreeParser treeParser = new CanonicalTreeParser();
-        final ObjectReader reader = repository.newObjectReader();
-        treeParser.reset(reader, tree.getId());
-        reader.close();
-        walk.dispose();
-        walk.close();
-        return treeParser;
+        try {
```
rnveach: I prefer line separations between groups/ideas to make reading easier. Please add 1 line between variable declarations and try.
Please do this for all.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -86,19 +97,31 @@ private DiffParser() {
...
+            final RevTree tree = walk.parseTree(commit.getTree().getId());
+            returnValue = new CanonicalTreeParser();
+            final ObjectReader reader = repository.newObjectReader();
+            try {
+                returnValue.reset(reader, tree.getId());
+            }
+            finally {
+                reader.close();
+            }
+            walk.dispose();
```
rnveach: 1 separation between `try-finally` and next statement.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
rnveach: It seems this is coming from IntelliJ and as in main repository, we have it disabled. See https://github.com/checkstyle/checkstyle/blob/7ba18e5138cdb800243b0a7dd15de3bb3741e395/config/intellij-idea-inspections.xml#L2140 .
Please confirm you use the editor IntelliJ and I will copy the `intellij-idea-inspections` from their to regression-tool.

Luolc: Yeah I am using Intellij. I would disable this inspection.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java

src/main/java/com/github/checkstyle/regression/git/DiffParser.java

src/main/java/com/github/checkstyle/regression/git/DiffParser.java

Issue #30, opened 2017-06-15
Add IntelliJ inspection file


rnveach:
Identified at https://github.com/checkstyle/regression-tool/pull/27#discussion_r122016862 ,

Main programmer uses IntelliJ, so we should incorporate Checkstyle's inspection file into this project.
We will do an 'as-is' copy for now.

rnveach:
Fix was merged

PR #31, opened 2017-06-15
Issue #30: added IntelliJ inspection files

rnveach:
Issue #30 

Modified `intellij-idea-inspection-scope.xml`  to have project's name and removed excess excludes that we won't have.

I don't use IntelliJ much and can't confirm right now if anything else is needed. I don't see any other mentions for it in Checkstyle's sources.
@Luolc Does this work for you, if you have time to test?
Otherwise I will try and confirm tonight.


Luolc:
@rnveach It works well on my Intellij.

Issue #5, opened 2017-06-05
Non-Default Configuration Testing


rnveach:
We need to go beyond default module testing and dive changing property values.
Reason can be seen at https://github.com/checkstyle/checkstyle/issues/3656#issuecomment-267135176 .

We can already identify properties, but how are going to decide what values to set? Booleans are easy since they have 2 values, but what about regular expression strings or files.
Should we hardcode some possible values and randomize the rest? Can we scan tests for property values?

How will testing multiple versions of the same module be handled for reporting as we need to know the results for the specific version of the module being checked.
If we build multiple configurations for multiple module versions, regression runtime will take longer.

PR #33, opened 2017-06-17
Issue #19: Module Component

Luolc:
#19 

This PR contains many things. Some changes and discussions may be better in other issue/PR. Free feel to point them out and create new issues for them.

1. I introduced [Immutables](https://immutables.github.io/). It is a apt tool to generate immutable java object with boilerplate methods, builder, and serialization/deserialization supports etc. Our POJOs might become huge in the future(i.e. `ModuleInfo` would contains many fields). So I think it is a good choice to introduce such library. FYI: libraries with similar function: [AutoValue](https://github.com/google/auto/tree/master/value), [FreeBuilder](https://github.com/google/FreeBuilder)
2. Currently I created the hardcode module info in json format. https://github.com/Luolc/regression-tool/blob/ee1fa4767cd6237f93683685d9e9c3a71a3ffced/src/main/resources/checkstyle_modules.json. I used the hardcode map in `PackageObjectFactory` and write a code snippet in `PackageObjectFactoryTest` to generate it. Contains the following info: name, parent, packageName, hierarchies, interfaces. But hierarchies, interfaces haven't been used in code yet.
3. I used [Gson](https://github.com/google/gson) to handle the hardcode info. In fact, the json is in a file so that when we dynamically generate the json in the future, the production code won't change a lot.
4. I changed the order of dependencies in `pom.xml`, splitting main and test dependencies. Maybe it is better to do that in a separate PR.
5. `pom.xml` was changed a bit to exclude the generated files from coverage check.

### DIFF
--- a/src/main/java/com/github/checkstyle/regression/module/ModuleCollector.java
+++ b/src/main/java/com/github/checkstyle/regression/module/ModuleCollector.java
@@ -1,0 +1,63 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.module;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import com.github.checkstyle.regression.data.GitChange;
+import com.github.checkstyle.regression.data.ModifiableModuleInfo;
+import com.github.checkstyle.regression.data.ModuleExtractInfo;
+import com.github.checkstyle.regression.data.ModuleInfo;
+
+/**
+ * Collects all the necessary information for the generation, in module level.
+ * @author LuoLiangchen
+ */
+public final class ModuleCollector {
+    /** Prevents instantiation. */
+    private ModuleCollector() {
+    }
+
+    /**
+     * Generates the module information from a list of changes.
+     * @param changes the changes source
+     * @return the module information generated from the given changes
+     */
+    public static List<ModuleInfo> generate(List<GitChange> changes) {
+        final Map<String, ModifiableModuleInfo> moduleInfos = new LinkedHashMap<>();
+
+        for (GitChange change : changes) {
+            if (ModuleUtils.isCheckstyleModule(change)) {
+                final String fullName = ModuleUtils.convertJavaSourceChangeToFullName(change);
+                final ModuleExtractInfo extractInfo = ModuleUtils.getModuleExtractInfo(fullName);
+                final ModifiableModuleInfo moduleInfo = ModifiableModuleInfo.create()
+                        .setModuleExtractInfo(extractInfo);
+                moduleInfos.put(extractInfo.fullName(), moduleInfo);
+            }
+        }
+
+        return moduleInfos.values().stream()
+                .map(ModifiableModuleInfo::toImmutable)
+                .collect(Collectors.toList());
+    }
+}
--- a/src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
+++ b/src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
@@ -1,0 +1,191 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.module;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import com.github.checkstyle.regression.data.GitChange;
+import com.github.checkstyle.regression.data.ModuleExtractInfo;
+
+/**
+ * Contains utility methods related to checkstyle module.
+ * @author LuoLiangchen
+ */
+public final class ModuleUtils {
+    /** The compiled regex pattern of the path of Java source files. */
+    private static final Pattern JAVA_SOURCE_PARTTEN =
+            Pattern.compile("src/(main|test)/java/(.+)\\.java");
+
+    /** The postfix of a test of a checkstyle module. */
+    private static final String TEST_POSTFIX = "Test";
+
+    /** The map of full qualified name to module extract info. */
+    private static final Map<String, ModuleExtractInfo> NAME_TO_MODULE_EXTRACT_INFO =
+            new HashMap<>();
+
+    /** Prevents instantiation. */
+    private ModuleUtils() {
+    }
+
+    /**
+     * Sets the map of full qualified name to module extract info with the given map.
+     * @param map the given map
+     */
+    public static void setNameToModuleExtractInfo(Map<String, ModuleExtractInfo> map) {
+        NAME_TO_MODULE_EXTRACT_INFO.clear();
+        NAME_TO_MODULE_EXTRACT_INFO.putAll(map);
+    }
+
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a checkstyle module.
+     * Changes of checkstyle modules are Java main source files, of which full
+     * qualified names are in the extract module info map.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a checkstyle module
+     */
+    public static boolean isCheckstyleModule(GitChange change) {
+        final boolean returnValue;
+        if (isJavaMainSource(change)) {
+            final String fullName = convertJavaSourceChangeToFullName(change);
+            returnValue = NAME_TO_MODULE_EXTRACT_INFO.containsKey(fullName);
+        }
+        else {
+            returnValue = false;
+        }
+        return returnValue;
+    }
+
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a checkstyle utility class.
+     * Changes of checkstyle utility classes are Java main source files, of which full
+     * qualified names are not in the extract module info map.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a checkstyle utility class.
+     */
+    public static boolean isCheckstyleUtility(GitChange change) {
+        final boolean returnValue;
+        if (isJavaMainSource(change)) {
+            final String fullName = convertJavaSourceChangeToFullName(change);
+            returnValue = !NAME_TO_MODULE_EXTRACT_INFO.containsKey(fullName);
+        }
+        else {
+            returnValue = false;
+        }
+        return returnValue;
+    }
+
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a test of checkstyle module.
+     * Changes of checkstyle module tests are Java test source files, of which full
+     * qualified names end with "Test" and the full names of corresponding modules are
+     * in the extract module info map.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a test of checkstyle module
+     */
+    public static boolean isCheckstyleModuleTest(GitChange change) {
+        final boolean returnValue;
+        if (isJavaTestSource(change)) {
+            final String fullName = convertJavaSourceChangeToFullName(change);
+            if (fullName.endsWith(TEST_POSTFIX)) {
+                returnValue = NAME_TO_MODULE_EXTRACT_INFO.containsKey(
+                        fullName.substring(0, fullName.length() - TEST_POSTFIX.length()));
+            }
+            else {
+                returnValue = false;
+            }
+        }
+        else {
+            returnValue = false;
+        }
+        return returnValue;
+    }
+
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a Java main source file.
+     * Changes of Java main source files are which have path matching
+     * {@code JAVA_SOURCE_PARTTEN} and in "main" directory.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a Java main source file.
+     */
+    private static boolean isJavaMainSource(GitChange change) {
+        final boolean returnValue;
+        final Matcher matcher = JAVA_SOURCE_PARTTEN.matcher(change.getPath());
+        if (matcher.find()) {
+            returnValue = "main".equals(matcher.group(1));
+        }
+        else {
+            returnValue = false;
+        }
+        return returnValue;
+    }
+
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a Java test source file.
+     * Changes of Java test source files are which have path matching
+     * {@code JAVA_SOURCE_PARTTEN} and in "test" directory.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a Java test source file.
+     */
+    private static boolean isJavaTestSource(GitChange change) {
+        final boolean returnValue;
+        final Matcher matcher = JAVA_SOURCE_PARTTEN.matcher(change.getPath());
+        if (matcher.find()) {
+            returnValue = "test".equals(matcher.group(1));
+        }
+        else {
+            returnValue = false;
+        }
+        return returnValue;
+    }
+
+    /**
+     * Gets the module extract info from the given full qualified name.
+     * @param fullName the given full qualified name
+     * @return the module extract info got from the given full qualified name
+     */
+    public static ModuleExtractInfo getModuleExtractInfo(String fullName) {
+        return NAME_TO_MODULE_EXTRACT_INFO.get(fullName);
+    }
+
+    /**
+     * Converts a change of Java source file to its full qualified name.
+     * @param change the change instance of Java source file
+     * @return the corresponding full qualified name
+     */
+    public static String convertJavaSourceChangeToFullName(GitChange change) {
+        return Arrays.stream(JAVA_SOURCE_PARTTEN.matcher(change.getPath())
+                .replaceAll("$2").split("/")).collect(Collectors.joining("."));
+    }
+}
--- a/src/test/java/com/github/checkstyle/regression/module/ModuleUtilsTest.java
+++ b/src/test/java/com/github/checkstyle/regression/module/ModuleUtilsTest.java
@@ -1,0 +1,98 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.module;
+
+import static com.github.checkstyle.regression.internal.TestUtils.assertUtilsClassHasPrivateConstructor;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.util.Map;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import com.github.checkstyle.regression.data.GitChange;
+import com.github.checkstyle.regression.data.ImmutableModuleExtractInfo;
+import com.github.checkstyle.regression.data.ModuleExtractInfo;
+import com.github.checkstyle.regression.extract.ExtractInfoProcessor;
+
+public class ModuleUtilsTest {
+    private static final String BASE_PACKAGE = "com.puppycrawl.tools.checkstyle";
+
+    private static final String JAVA_MAIN_SOURCE_PREFIX =
+            "src/main/java/com/puppycrawl/tools/checkstyle/";
+
+    @Before
+    public void setUp() {
+        final InputStream is = ExtractInfoProcessor.class.getClassLoader()
+                .getResourceAsStream("checkstyle_modules.json");
+        final InputStreamReader reader = new InputStreamReader(is, Charset.forName("UTF-8"));
+        final Map<String, ModuleExtractInfo> map =
+                ExtractInfoProcessor.getModuleExtractInfosFromReader(reader);
+        ModuleUtils.setNameToModuleExtractInfo(map);
+    }
+
+    @Test
+    public void testIsProperUtilsClass() throws Exception {
+        assertUtilsClassHasPrivateConstructor(ModuleUtils.class);
+    }
+
+    @Test
+    public void testIsCheckstyleModule() {
+        final GitChange change = new GitChange(
+                JAVA_MAIN_SOURCE_PREFIX + "checks/coding/EmptyStatementCheck.java");
+        final boolean result = ModuleUtils.isCheckstyleModule(change);
+        assertTrue("EmptyStatementCheck should be considered as a checkstyle module", result);
+    }
+
+    @Test
+    public void testIsCheckstyleModuleNonModule() {
+        final GitChange change = new GitChange(
+                JAVA_MAIN_SOURCE_PREFIX + "PackageObjectFactory.java");
+        final boolean result = ModuleUtils.isCheckstyleModule(change);
+        assertFalse("PackageObjectFactory should not be considered as a checkstyle module", result);
+    }
+
+    @Test
+    public void testIsCheckstyleModuleNonMainFile() {
+        final boolean result =
+                ModuleUtils.isCheckstyleModule(new GitChange("src/test/java/foo/Foo.java"));
+        assertFalse("Non main file should not be consideres as a checkstyle module", result);
+    }
+
+    @Test
+    public void testConvertModuleChangeToExtractInfo() {
+        final GitChange change = new GitChange(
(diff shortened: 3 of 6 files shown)


config/import-control.xml
```diff
@@ -7,7 +7,18 @@
...
     <subpackage name="git">
         <allow pkg="org.eclipse.jgit"/>
     </subpackage>
+    <subpackage name="module">
+        <allow pkg="com.google.gson"/>
+        <!-- Keep this until we don't need hardcode ModuleInfo map. -->
+        <allow class="java.nio.charset.Charset"/>
+    </subpackage>
+    <subpackage name="util">
+        <allow pkg="com.google.gson"/>
```
rnveach: I am fine with `util` package, but it seems to me we are slightly breaking component design by having a component specific import outside it's package.
`gson` will never be used by another component besides `module`, right?
Should this `util` package be located inside `module`?

Looking at the `ModuleUtil`, it seems like we are breaking apart the one `module` class in the wrong way.
Anything releated to the file reading should be in a class by itself, either in a `util` package or not.
Either everything should be in the main class for the component, or split by main and change parsing. This sound good?

src/main/java/com/github/checkstyle/regression/module/ConfigGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+    public static List<ConfigNode> generateConfigNodes(List<GitChange> changes) {
+        final List<ModuleInfo> modules = changes.stream()
+                .filter(ModuleUtils::isCheckstyleModule)
+                .map(ModuleUtils::convertModuleChangeToModuleInfo)
+                .collect(Collectors.toList());
+        return modules.stream()
+                .map(module -> {
+                    final ConfigNode.Parent parent;
+                    if (module.hasCheckerParent()) {
+                        parent = ConfigNode.Parent.CHECKER;
```
rnveach: We define `parent` in JSON file. We can't assume there will only be these 2 in a PR.
Parent needs to be auto-detected and auto-built just like module and it's properties.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,133 @@
...
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+
+/**
+ * Contains utility methods related to checkstyle module.
+ * @author LuoLiangchen
+ */
+public final class ModuleUtils {
+    /** The compiled regex pattern of the path of Java main source files. */
+    private static final Pattern MAIN_SOURCE_PATTERN = Pattern.compile("src/main/java/(.+)\\.java");
```
rnveach: Will windows require `\`? If so, we may need `[\\/]` like our suppressions.
I will look into this and get back.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,133 @@
...
+            try {
+                modules = gson.fromJson(reader, new TypeToken<List<ModuleInfo>>() {
+                }.getType());
+            }
+            finally {
+                reader.close();
+            }
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException("Failed when loaing hardcode information", ex);
```
rnveach: `hardcode information` => `hardcode checkstyle module information`
Best to be as specific as possible.

src/main/java/com/github/checkstyle/regression/util/GsonProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final Gson GSON;
+
+    /** Prevents instantiation. */
+    private GsonProvider() {
+    }
+
+    static {
+        final GsonBuilder gsonBuilder = new GsonBuilder();
+        for (TypeAdapterFactory factory : ServiceLoader.load(TypeAdapterFactory.class)) {
+            gsonBuilder.registerTypeAdapterFactory(factory);
```
rnveach: What are these 2 lines doing?
Can and should this be embedded with the other file stuff?

src/test/java/com/github/checkstyle/regression/module/ConfigGeneratorTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final String JAVA_MAIN_SOURCE_PREFIX =
+            "src/main/java/com/puppycrawl/tools/checkstyle/";
+
+    @Test
+    public void testIsProperUtilsClass() throws Exception {
+        assertUtilsClassHasPrivateConstructor(ConfigGenerator.class);
+    }
+
+    @Test
+    public void testGenerateConfigNodes() {
```
rnveach: Split this into 2 tests. Invalid and valid changes. (valid = recognize as checkstyle module)
Add a new invalid location inside `JAVA_MAIN_SOURCE_PREFIX` and inside `src\main\java` but both not recognized as a module. This is an example of a checkstyle utility class or a brand new module that we didn't recognize.

src/test/java/com/github/checkstyle/regression/module/ModuleUtilsTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    public void testGetNameToModuleInfoFromInputStreamWithException() throws Exception {
+        final Method method = ModuleUtils.class.getDeclaredMethod(
+                "getNameToModuleInfoFromInputStream", InputStream.class);
+        method.setAccessible(true);
+        final InputStream is = spy(ModuleUtils.class.getClassLoader()
+                .getResourceAsStream("checkstyle_modules.json"));
+        doThrow(IOException.class).when(is).close();
+        try {
```
rnveach: empty line before try.

src/test/java/com/github/checkstyle/regression/module/ConfigGeneratorTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+
+import org.junit.Test;
+
+import com.github.checkstyle.regression.data.ConfigNode;
+import com.github.checkstyle.regression.data.GitChange;
+import com.github.checkstyle.regression.data.ImmutableConfigNode;
+
+public class ConfigGeneratorTest {
+    private static final String BASE_PACKAGE = "com.puppycrawl.tools.checkstyle";
```
rnveach: Field is never used.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
rnveach: `ConfigGeneratorTest` fails when I change all file paths from `/` to `\\`.
I used another utility I wrote with jgit to see how it returns files and folders. `DiffEntry`'s paths are returned with `/` in Windows.

So it seems this won't be an issue and can be ignored for now.

config/import-control.xml
```diff
@@ -7,7 +7,18 @@
     <allow pkg="java.io"/>
     <allow pkg="java.util"/>
     <allow pkg="com.github.checkstyle.regression.data"/>
+    <allow pkg="com.github.checkstyle.regression.util"/>
+    <allow pkg="org.immutables"/>
```
rnveach: @Luolc I asked romani his opinion on this, and he isn't a fan of "magic" builders either.
Since this isn't the main repository, we could keep this here if we really wanted to.

My only concern right now is code coverage as `ModuleInfo` isn't showing any, even for `hasCheckerParent` which is a branch because of the equality test.

src/main/java/com/github/checkstyle/regression/module/ConfigGenerator.java
Luolc: What value can `parent` be? Actually I felt there are only these two before... I also generate the json file by a simple if-else(If the module is a subclass of AbstractChecker, then TreeWalker, otherwise Checker)

config/import-control.xml
Luolc: done.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/util/GsonProvider.java
Luolc: This is for registering the GsonTypeAdapter generated by immutables. It is sort of initialization of Gson. 
FYI: https://immutables.github.io/json.html#type-adapter-registration

src/main/java/com/github/checkstyle/regression/module/ConfigGenerator.java
rnveach: Technically any value. The only requirement in the JSON right now is it is a string.
Checkstyle only has these 2 modules with siblings right now, but there is nothing stopping us from adding new ones in the future. Your code has to be able to work with anything in the PR.

One example of this changing is the possibility of a new RootModule for multi-threaded mode, but this might require there be many possible parents instead of just 1.

> I also generate the json file by a simple if-else

This is fine for our current purposes. When we actually write this code and merge it somewhere, it may have to support the possibility of new parents.

src/test/java/com/github/checkstyle/regression/module/ModuleUtilsTest.java
Luolc: Done.

src/test/java/com/github/checkstyle/regression/module/ConfigGeneratorTest.java
Luolc: Removed.

src/test/java/com/github/checkstyle/regression/module/ConfigGeneratorTest.java
Luolc: I split the current UT into two. Not very clear about "new invalid" mean. And what does "inside" mean?
Added UT:
```Java
@Test
    public void testGenerateConfigNodesForInvalidChanges() {
        final List<GitChange> changes = new LinkedList<>();
        changes.add(new GitChange(
                JAVA_MAIN_SOURCE_PREFIX + "PackageObjectFactory.java"));
        changes.add(new GitChange("foo/A.java"));
        final List<ConfigNode> configNodes = ConfigGenerator.generateConfigNodes(changes);
        assertEquals(0, configNodes.size());
    }
```
> a new invalid location inside `JAVA_MAIN_SOURCE_PREFIX`

I guess ` changes.add(new GitChange(JAVA_MAIN_SOURCE_PREFIX + "PackageObjectFactory.java"));` is this situation.

> inside `src\main\java`

Do you mean smth like `src/main/java/Foo.java`?

src/main/java/com/github/checkstyle/regression/module/ConfigGenerator.java
Luolc: OK. Got it. I will update this to fit any strings.

config/import-control.xml
```diff
@@ -7,7 +7,21 @@
...
     </subpackage>
+
+    <subpackage name="module">
+        <allow pkg="com.google.gson"/>
+        <!-- Keep this until we don't need hardcode ModuleInfo map. -->
+        <allow class="java.nio.charset.Charset"/>
+    </subpackage>
+
+    <subpackage name="util">
+        <allow pkg="com.google.gson"/>
```
rnveach: I'm not sure what happened to my original post on this as I can't find it anymore.

I am fine with `util` package, but it seems to me we are breaking component design by using component specific `com.google.gson` outside of `module` component, which is the only component that will use it right?
Should this be a `util` inside module?

It also seems like `ModuleUtils` is trying to do too much. It does json reading and other various things.
I think we should have one utility for JSON, including code in `GsonProvider`, and have another class that is what util module needs.

src/test/java/com/github/checkstyle/regression/module/ConfigGeneratorTest.java
rnveach: >  "new invalid"  mean

I just meant a new location that you hadn't used before.

> "inside" mean

A file inside that folder location.

> I guess changes.add(new GitChange(JAVA_MAIN_SOURCE_PREFIX + "PackageObjectFactory.java")); is this situation.

Yes.

> Do you mean smth like src/main/java/Foo.java?

Yes.

src/main/java/com/github/checkstyle/regression/module/ConfigGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+    /** Prevents instantiation. */
+    private ConfigGenerator() {
+    }
+
+    /**
+     * Generates the regression config nodes from a list of changes.
+     * @param changes the changes source
+     * @return the regression config nodes generated from the given changes
+     */
+    public static List<ConfigNode> generateConfigNodes(List<GitChange> changes) {
```
Luolc: I think it depends on the duty scope of `configuration`. I was thinking that `configuration` would only get direct data describing the XML.(module name, property name, value etc.) If we are going to pass more data to `configuration`, then how detailed it would be?

rnveach: The only duty of `configuration` is build the configuration file. There are no specifics on the type of data it should retrieve.
Someone has to build the configuration, and someone has to generate the list of different modules to build based on properties and such. It's not a concern who does which, but do we need an intermediate data type before the `direct data describing the XML`.

> I was thinking that configuration would only get direct data describing the XML.(

My reasoning against this is "what if we have to build a new component between module and configuration". Can this new component deal with the design structure of the final configuration?
Would it just want to see a single module as a whole?

We can hold this off for now, but I am just worried about if this structure will be good for https://github.com/checkstyle/regression-tool/issues/5 as we need to create multiple module configurations for each set of properties.
Which is why I think the answer to `How will you maintain list of properties and values to cycle through?` is an important question.

config/import-control.xml
Luolc: I believe there would be many fields in `ModuleInfo` in the future. 
If we use constructor to pass these parameters, it would be ugly. 《Effective Java》 said it is better to use builder style if constructor has more than three parameters.
Another annoying thing is the `equals` method. We would have a lot of `equals` asserting in UTs. It is hard to appease codecov if we write this manually. Supposed there are 5 fields then the `equals` method would have 32 branches(2^5). That's horrible. If we write a `equals` method in UTs, yeah that could be a workaround, be a little bit strange. And it is also bad that when we add or change some fields, then we have to update the utility `equals` in UTs.

So I think it is not bad if an annotation processing tool could generate these for us. We could then focus on more important things but not writing builders, equals and thinking how to appease codecov.

rnveach: > Supposed there are 5 fields then the equals method would have 32 branches(2^5). That's horrible. 

I assume we got around this in main project using `equalsverifier` dependency.

> So I think it is not bad if an annotation processing tool could generate these for us

Let's keep `org.immutables` than, but how about we require no custom methods in the interface and require them to be in `utility` classes so we can keep code coverage on them? I dislike the idea of having code we can't be sure we are covering.

rnveach:
@Luolc ping
Its been 2 days without any acitivity here. Is everything going ok?

Luolc:
@rnveach 
1. I added the middleware `ModuleUpdate`. What's your opinion of its current contents? I am not very confident of it and need your suggestion.
2. I haven't handled the location gson util and splitting gson part from `ModuleUtils` yet. Maybe focus on them sooner.

rnveach:
> ModuleUpdate

I don't think this class can be fully immutable. How are you going to combine different changes in separate `GitChange`? Will you recreate the entire class to add a new change?
The point of the `middleware` is that it will be in flux while we parse all the git changes. We have to be able to add everything we need to this one class. I am not expecting `GitChange` to go to next component. `module` will do everything with it now.



Here is what I recommend:
1) Rename `ModuleInfo` to `ModuleExtractInfo` and `ModuleUpdate` to `ModuleInfo` and `ModuleUpdateCollector` to `ModuleCollector`.
2) `ModuleInfo` should be a non-immutable class and contain: name (module), module extract, list of unique property names and values (ignore properties for now).
3) Rename `generateModuleUpdates` to `generate`.
4) Inside `parse`:
a) create map of module name and `ModuleInfo`.
b) for each change,
i) If file is module, add module to map if doesn't exist
ii) If file is test module, add module to map if doesn't exist (eventually expanding this to grab properties)
c) return changes as list.

When we go to next component, it will take `ModuleInfo` and start creating the configuration from it.
Eventually, if no properties are found or not all, we will generate our own (some how).

Does this make sense?

Luolc:
> If file is test module, add module to map if doesn't exist

What if in a PR there is only one test file changed? In this case we don't need regression. We have many issues that only need to change test scope without touching production code.

rnveach:
> IMO, we should go through the production code changes first, adding some modules to the map(direct change now, and gain related modules from utility changes in the future). Then, go through the test changes. If the map contains the corresponding module already, then process the test change, otherwise just ignore it.

Agreed.

Luolc:
> Rename ModuleInfo to ModuleExtractInfo and ModuleUpdate to ModuleInfo and ModuleUpdateCollector to ModuleCollector.
ModuleInfo should be a non-immutable class and contain: name (module), module extract, list of unique property names and values (ignore properties for now).
Rename generateModuleUpdates to generate.
Inside parse:
a) create map of module name and ModuleInfo.
b) for each change,
i) If file is module, add module to map if doesn't exist
ii) If file is test module, add module to map if doesn't exist (eventually expanding this to grab properties)
c) return changes as list.

Updated according to the suggestions. Please have a check.

config/import-control.xml
rnveach: @Luolc Please see this.

config/import-control.xml
rnveach: Keeping them as abstract classes seems to keep them requiring coverage, so lets stick with this.

src/main/java/com/github/checkstyle/regression/util/GsonProvider.java
rnveach: As discussed below, lets keep all `gson` stuff together in one class.

config/import-control.xml
rnveach: One small update, lets move all `gson` code to new component in https://github.com/checkstyle/regression-tool/issues/37 .
This component will now receive the collection of `ModuleExtractInfo`s.

rnveach: Lets move all gson code to new component in #37 .
Either this component will now receive the collection of ModuleExtractInfos, or a utility class will and keep it as global information.

src/main/resources/checkstyle_modules.json
```diff
@@ -0,0 +1,2340 @@
+[
+  {
+    "name": "Checker",
```
rnveach: @Luolc See https://github.com/checkstyle/regression-tool/issues/37 .
Lets remove this file from `main` and use it only in test for our UTs.
It will only be used in #37 .

rnveach:
@Luolc Everything is looking good.

Just to summarize:

1)
Lets move all gson code to new component in #37 .
Either this component will now receive the collection of ModuleExtractInfos, or a utility class will and keep it as global information.

2)
Lets remove `checkstyle_modules.json` from `main` and use it only in test for our UTs.
It will only be used in #37 .

3)
Remove commented code. If it is needed for future items, keep them as a copy somewhere.

Coverage can be less than 100% for now to get this merged faster.

Luolc:
1. Is this component only for json related code and named as `json`?

Don't have questions on others

rnveach:
> Is this component only for json related code and named as json?

Answered in issue #37 .

Luolc:
> Split dependency re-ordering into minor commit

Split to #38 

Then changes about adding objects to `data` are split to #39 . It is blocked by #38 , we have to finish #38 and then rebase and go on.

> Lets move all gson code to new component in #37

In #40 , also, blocked by #38 and #39 

Other changes are also done and pushed, but need these PRs above are merged and rebase them.

rnveach:
Please rebase to remove conflicts. Only changes for main component should remain.

Luolc:
Rebased and pushed. I am adding the assert messages in test now.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,172 @@
...
+            returnValue = false;
+        }
+        return returnValue;
+    }
+
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a Java main source file.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
```
rnveach: duplicating main description in return tag is bad javadoc design. You are not giving reader anything new to read.
Can you try to fix these and others by making each unique and different information?

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a test of checkstyle module.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a test of checkstyle module
+     */
+    public static boolean isCheckstyleModuleTest(GitChange change) {
+        final boolean returnValue;
+        if (JAVA_SOURCE_PARTTEN.matcher(change.getPath()).find()) {
```
rnveach: This is `true` for test and main. It should only be accepting test. Make a `isJavaTestSource` and use that.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /** Prevents instantiation. */
+    private ModuleUtils() {
+    }
+
+    /**
+     * Sets the map of full qualified name to module extract info with the given map.
+     * @param map the given map
+     */
+    public static void setNameToModuleExtratInfo(Map<String, ModuleExtractInfo> map) {
```
rnveach: `Extrat` => `Extract`

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,172 @@
...
+        final String fullName = convertJavaSourcePathToFullName(change.getPath());
+        return NAME_TO_MODULE_EXTRAT_INFO.get(fullName);
+    }
+
+    /**
+     * Converts a {@link GitChange} of test to the full name of corresponding checkstyle module.
+     * @param change the change to convert
+     * @return the full name of corresponding checkstyle module
+     */
+    public static String convertModuleTestChangeToModuleFullName(GitChange change) {
```
rnveach: There is no test that change is in test and no test that `testName` ends with `Test` like `isCheckstyleModuleTest`.

src/main/java/com/github/checkstyle/regression/module/ModuleCollector.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Generates the module information from a list of changes.
+     * @param changes the changes source
+     * @return the module information generated from the given changes
+     */
+    public static List<ModuleInfo> generate(List<GitChange> changes) {
+        final Map<String, ModifiableModuleInfo> moduleInfos = new LinkedHashMap<>();
+        for (GitChange change : changes) {
```
rnveach: empty line before `for`.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,172 @@
...
+        final String testName = convertJavaSourcePathToFullName(change.getPath());
+        return testName.substring(0, testName.length() - TEST_POSTFIX.length());
+    }
+
+    /**
+     * Converts a path of Java source file to its full qualified name.
+     * @param path the path of Java source file
+     * @return the corresponding full qualified name
+     */
+    private static String convertJavaSourcePathToFullName(String path) {
```
rnveach: `convertModuleChangeToExtractInfo` and `convertModuleTestChangeToModuleFullName` don't really need `GitChange` but you leave `convertJavaSourcePathToFullName` as `String` even though everywhere that uses it has `GitChange` available.

Change this parameter to GitChange.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,172 @@
...
+public final class ModuleUtils {
+    /** The compiled regex pattern of the path of Java source files. */
+    private static final Pattern JAVA_SOURCE_PARTTEN =
+            Pattern.compile("src/(main|test)/java/(.+)\\.java");
+
+    /** The postfix of a test of a checkstyle module. */
+    private static final String TEST_POSTFIX = "Test";
+
+    /** The map of full qualified name to module extract info. */
+    private static final Map<String, ModuleExtractInfo> NAME_TO_MODULE_EXTRAT_INFO =
```
rnveach: `EXTRAT` => `EXTRACT`
Check for this misspelling everywhere.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
rnveach: These 2 convert methods (`convertModuleTestChangeToModuleFullName` and `convertModuleChangeToExtractInfo`) make me feel like you are trying to put too much into this class that may not be best as a utility method.
Should we make `convertJavaSourcePathToFullName` public and have the main component do the work instead of these methods?

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/ModuleCollector.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: I add some info in each main description. Refers to the similar utility methods in `ModuleReflectionUtils` in checkstyle project.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: The main concern is `NAME_TO_MODULE_EXTRACT_INFO` map. It should be placed in `ModuleUtils` since `isCheckstyleModule`, `isCheckstyleUtility`, `isCheckstyleModuleTest` all need it.

`convertModuleTestChangeToModuleFullName` is OK to move out from `ModuleUtils` now. But  `convertModuleChangeToExtractInfo` could not, since we need access `NAME_TO_MODULE_EXTRACT_INFO` to get the converted info. I don't think we should place `NAME_TO_MODULE_EXTRACT_INFO` somewhere else.

My opinion now is: keeping these two methods here, then adding a if-condition at the beginning of each, throwing a `IllegalArgumentException` if the parameter is not valid(not checkstyle module, or not checkstyle module test)

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * Checks whether the corresponding file of a change may be considered as
+     * a Java test source file.
+     * Changes of Java test source files are which have path matching
+     * {@code JAVA_SOURCE_PARTTEN} and in "test" directory.
+     * @param change change to check
+     * @return true if the corresponding file of a change may be considered as
+     *      a Java test source file.
+     */
+    private static boolean isJavaTestSource(GitChange change) {
```
rnveach: One last thing, please group `isJava...` methods together. Do the same for `isCheckstyle...` methods.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
rnveach: These methods just seem too specific to be an actual utility method. 1 is unused, so it could be one issue for why I can't see a good way to fix it right now.

Yes, the map should stay.
I was thinking not to move them out but having simplier methods like `getModuleInfo` (take full module name as string parameter) so you don't have to move it out. The only issue for the test method would be `TEST_POSTFIX` which I don't see how you can get around for now.

Otherwise, you can throw the exception, but the test one also needs to make sure it ends with `Test`.
Usage of this method should never throw the exception, so you have to be real careful in it's use. I might make an issue to look at it later when we implement more in the module component.

Luolc: > I was thinking not to move them out but having simplier methods like getModuleInfo (take full module name as string parameter)

I am fine with this.

Since actually we haven't use `convertModuleTestChangeToModuleFullName` now. How about just remove it now and decide it later when we have more codes of handling test files later. Maybe it would be clearer of how to handle this at that time.

rnveach: Yes, just remove it than.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/ModuleUtils.java
Luolc: Done.

rnveach:
Merged.

PR #38, opened 2017-06-26
minor: introduce Immutables, Gson and re-order the dependencies

Luolc:

Issue #37, opened 2017-06-25
Extract Component - Generate Module Extract list


rnveach:
Taken from discussions in https://github.com/checkstyle/regression-tool/issues/29 ,

We need to generate a list of modules, properties, and hierarchy from checkstyle PR for our analysis and to identify full regression model needed.

This the list of information we need right now:

````
-- Module Information

Package
Class Name
Class hierarchy: Can contain multiple classes. If AbstractSuperCheck is modified, we need to determine that SuperCloneCheck and SuperFinalizeCheck need regression. Must contain full class path.
Interfaces: Same reason as class heirarchy.
Parent Module: Checker or TreeWalker
List of properties

-- Property Information

Name
Type
Default Value
Acceptable Tokens (only for Java and Javadoc checks)
````

Based on discussions at https://github.com/checkstyle/regression-tool/issues/19#issuecomment-308527052 and in issue:

As of right now, this probably has to be loaded as a file from a separate maven process.

> invoke some simple program in our PR copy of Checkstyle through maven and have it generate the results for us that we need. It will pass the information back to our regression-tool's JVM by a file. We don't need the actual class in memory, we just need it's meta-information.

So we don't have to inject `gson` dependency into Checkstyle too, I think we should just write out file by hand.
It is ok if we can't cover external maven command by coverage or UT for now.

This is blocked by https://github.com/checkstyle/regression-tool/issues/19 .

rnveach:
See https://github.com/checkstyle/checkstyle/blob/0b6721b6b4ad77115cd443f9a7bddf973c2ab224/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java and https://github.com/checkstyle/checkstyle/blob/0b6721b6b4ad77115cd443f9a7bddf973c2ab224/src/test/java/com/puppycrawl/tools/checkstyle/internal/CheckUtil.java on how to get this information we need.

Modules: https://github.com/checkstyle/checkstyle/blob/0b6721b6b4ad77115cd443f9a7bddf973c2ab224/src/test/java/com/puppycrawl/tools/checkstyle/internal/CheckUtil.java#L151
Properties: https://github.com/checkstyle/checkstyle/blob/0b6721b6b4ad77115cd443f9a7bddf973c2ab224/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L881
[...]

PR #39, opened 2017-06-26
Issue #19: Add data object

Luolc:
Taken from #33 . This is blocked by #38

### DIFF
--- a/config/findbugs-exclude.xml
+++ b/config/findbugs-exclude.xml
@@ -15,4 +15,9 @@
             <Bug pattern="BC_UNCONFIRMED_CAST_OF_RETURN_VALUE"/>
         </Or>
     </Match>
+    <Match>
+        <!-- This is a bug of immutables. hashCode() of a Modifiable class is dangerous. -->
+        <Class name="~.*\.Modifiable.*"/>
+        <Bug pattern="UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR"/>
+    </Match>
 </FindBugsFilter>
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -7,6 +7,13 @@
     <allow pkg="java.io"/>
     <allow pkg="java.util"/>
     <allow pkg="com.github.checkstyle.regression.data"/>
+    <allow pkg="org.immutables.value"/>
+
+    <subpackage name="data">
+        <!-- it is only allowed until https://github.com/checkstyle/checkstyle/issues/3492 -->
+        <allow pkg="org.immutables.gson"/>
+    </subpackage>
+
     <subpackage name="git">
         <allow pkg="org.eclipse.jgit"/>
     </subpackage>
--- a/pom.xml
+++ b/pom.xml
@@ -285,10 +285,23 @@
                                         <branchRate>0</branchRate>
                                         <lineRate>0</lineRate>
                                     </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.data.ModuleExtractInfo</pattern>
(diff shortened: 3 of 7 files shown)


config/import-control.xml
```diff
@@ -7,6 +7,8 @@
     <allow pkg="java.io"/>
     <allow pkg="java.util"/>
     <allow pkg="com.github.checkstyle.regression.data"/>
+    <allow pkg="org.immutables"/>
```
rnveach: Change this to `org.immutables.value`.

`org.immutables.gson` should only be specifically allowed for the class `ModuleExtractInfo` as this conflicts with https://github.com/checkstyle/regression-tool/issues/37#issuecomment-310950569 since gson is related to extract file.

> This component (extract component) ... should contain only code relating to everything needed to  ... read the Checkstyle extract file 

It would be even better if we could do away with this import and signify it as a `TypeAdapters` another way.

pom.xml
```diff
@@ -23,27 +23,25 @@
 
     <dependencies>
         <dependency>
-            <groupId>junit</groupId>
```
rnveach: Rebase and none of these dependency items should remain.

config/import-control.xml
rnveach: > should only be specifically allowed for the class 

This isn't allowed yet as it is still an open issue at https://github.com/checkstyle/checkstyle/issues/3492 .
Move this to `data` package and make a comment above line that it is only allowed until `https://github.com/checkstyle/checkstyle/issues/3492`, if there is no other way to signify it as a `TypeAdapters`.

pom.xml
Luolc: Done.

src/main/java/com/github/checkstyle/regression/data/ModuleInfo.java
```diff
@@ -0,0 +1,73 @@
...
+package com.github.checkstyle.regression.data;
+
+import java.util.List;
+
+import org.immutables.value.Value;
+
+/**
+ * Represents the related information of a checkstyle module.
+ * Including the extract module information, change of the module class, change
+ * of the corresponding test and changed of the utility classes that the module
```
rnveach: Sorry, I missed this.
There are no more change information.

config/findbugs-exclude.xml
```diff
@@ -15,4 +15,11 @@
             <Bug pattern="BC_UNCONFIRMED_CAST_OF_RETURN_VALUE"/>
         </Or>
     </Match>
+    <Match>
+        <Or>
+            <Class name="~.*\.GsonAdapters.*"/>
+            <Class name="~.*\.Immutable.*"/>
+            <Class name="~.*\.Modifiable.*"/>
```
rnveach: Shouldn't there be a `Bug pattern` in here? Aren't we just suppressing everything for the classes as they are written now?

src/main/java/com/github/checkstyle/regression/data/ModuleInfo.java
Luolc: Done.

config/findbugs-exclude.xml
Luolc: My fault. I added for ignoring during my test locally before.

I just have a check that `Modifiable*` class does have a bug pattern `UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR` and it is a bug of immutables lib. `hashCode` of `Modifiable*` class would raise a NPE if it contains null field.
But it won't affect us since we won't use any of it as map key or set value.

I updated the exclude file.

PR #40, opened 2017-06-26
Issue #37: json file processor

Luolc:
Issue #37

Taken from #33 , this is blocked by #38 and #39

### DIFF
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -14,6 +14,12 @@
         <allow pkg="org.immutables.gson"/>
     </subpackage>
 
+    <subpackage name="extract">
+        <allow pkg="com.google.gson"/>
+        <allow class="com.sun.tools.javac.util.ServiceLoader"/>
+        <allow class="java.lang.reflect.Type"/>
+    </subpackage>
+
     <subpackage name="git">
         <allow pkg="org.eclipse.jgit"/>
     </subpackage>
--- a/src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
+++ b/src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
@@ -1,0 +1,86 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.extract;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import com.github.checkstyle.regression.data.ModuleExtractInfo;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.sun.tools.javac.util.ServiceLoader;
+
+/**
+ * Processes the module extract info grabbed from checkstyle project.
+ * @author LuoLiangchen
+ */
+public final class ExtractInfoProcessor {
+    /** The {@link Gson} instance. */
+    private static final Gson GSON;
+
+    /** The type of {@code List<ModuleExtractInfo>}. */
+    private static final Type TYPE_EXTRACT_INFO_LIST = new TypeToken<List<ModuleExtractInfo>>() {
+    }.getType();
+
+    /** Prevents instantiation. */
+    private ExtractInfoProcessor() {
+    }
+
+    static {
+        final GsonBuilder gsonBuilder = new GsonBuilder();
+        for (TypeAdapterFactory factory : ServiceLoader.load(TypeAdapterFactory.class)) {
+            gsonBuilder.registerTypeAdapterFactory(factory);
+        }
+        GSON = gsonBuilder.create();
+    }
+
+    /**
+     * Gets the module extract info map from the given reader. Map key is the
+     * fully qualified module name.
+     * @param reader the given reader
+     * @return the full qualified name to module extract info map
+     */
+    public static Map<String, ModuleExtractInfo> getModuleExtractInfosFromReader(
+            Reader reader) {
+        final List<ModuleExtractInfo> modules;
+
+        try {
+            try {
+                modules = GSON.fromJson(reader, TYPE_EXTRACT_INFO_LIST);
+            }
+            finally {
+                reader.close();
+            }
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException(
+                    "Failed when loaing hardcoded checkstyle module information", ex);
+        }
+
+        return modules.stream().collect(Collectors.toMap(
+                ModuleExtractInfo::fullName, extractInfo -> extractInfo));
+    }
+}
--- a/src/main/java/com/github/checkstyle/regression/extract/package-info.java
+++ b/src/main/java/com/github/checkstyle/regression/extract/package-info.java
@@ -1,0 +1,24 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+/**
+ * Contains execution of checkstyle maven extract process and extract file processor.
+ * @author LuoLiangchen
+ */
+package com.github.checkstyle.regression.extract;
--- a/src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
+++ b/src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
@@ -1,0 +1,92 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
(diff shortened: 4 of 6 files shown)


src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+    public static Map<String, ModuleExtractInfo> getNameToModuleExtractInfoFromInputStream(
+            InputStream stream) {
+        final List<ModuleExtractInfo> modules;
+
+        try {
+            final InputStreamReader reader =
+                    new InputStreamReader(stream, Charset.forName("UTF-8"));
+
+            try {
+                modules = GSON.fromJson(reader, new TypeToken<List<ModuleExtractInfo>>() {
```
rnveach: Please move `new TypeToken` to a static final field.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+        GSON = gsonBuilder.create();
+    }
+
+    /**
+     * Gets the full qualified name to module extract info map from the given input stream.
+     * @param stream the given input stream
+     * @return the full qualified name to module extract info map
+     */
+    public static Map<String, ModuleExtractInfo> getNameToModuleExtractInfoFromInputStream(
+            InputStream stream) {
```
rnveach: `InputStream` is fine, but you take that and than wrap it around `InputStreamReader` for UTF 8 support.
Shouldn't class that creates the stream define what charset the file should support and not this method?
I think we should either remove `InputStreamReader` and stick with `InputStream` or change this parameter to `InputStreamReader`.

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
```diff
@@ -0,0 +1,66 @@
...
+        assertUtilsClassHasPrivateConstructor(ExtractInfoProcessor.class);
+    }
+
+    @Test
+    public void testGetNameToModuleInfoFromInputStream() {
+        final InputStream is = ExtractInfoProcessor.class.getClassLoader()
+                .getResourceAsStream("checkstyle_modules.json");
+        final Map<String, ModuleExtractInfo> map =
+                ExtractInfoProcessor.getNameToModuleExtractInfoFromInputStream(is);
+        assertEquals(165, map.size());
```
rnveach: Trim test `json` file to only include items needed for specific testing.
We don't need full copy of checkstyle repo for basic testing.
If we find something we need later in it, we will expand.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
Luolc: > change this parameter to InputStreamReader

Done.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
Luolc: Done.

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
Luolc: OK. And, IMO, it is better to create another json file only for `ExtractInfoProcessorTest`, with a name like `checkstyle_modules_sample.json`(and maybe in `resources/.../extract`).

config/import-control.xml
```diff
@@ -7,6 +7,15 @@
     <allow pkg="java.io"/>
     <allow pkg="java.util"/>
     <allow pkg="com.github.checkstyle.regression.data"/>
+    <allow pkg="org.immutables"/>
```
rnveach: This and other similar items should go away with rebase.

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    public void testGetNameToModuleInfoFromInputStream() {
+        final InputStream is = ExtractInfoProcessor.class.getClassLoader()
+                .getResourceAsStream("checkstyle_modules.json");
+        final InputStreamReader reader = new InputStreamReader(is, Charset.forName("UTF-8"));
+        final Map<String, ModuleExtractInfo> map =
+                ExtractInfoProcessor.getNameToModuleExtractInfoFromInputStream(reader);
+        assertEquals(165, map.size());
```
rnveach: Was listed as outdated but is still not done.

> it is better to create another json file only for ExtractInfoProcessorTest, with a name like checkstyle_modules_sample.json

I am fine with 1 or multiple json files. If we do multiple, they should be trimmed down to bare bones for test.

Luolc: Emm... now I decide to keep this json file name currently. And if we find it is necessary to split the json file when we are doing main code of module component, we do the split then.

Luolc:
Done: Rebased and json trimmed.

config/import-control.xml
```diff
@@ -14,6 +14,13 @@
         <allow pkg="org.immutables.gson"/>
     </subpackage>
 
+    <subpackage name="extract">
+        <allow pkg="com.google.gson"/>
+        <allow class="com.sun.tools.javac.util.ServiceLoader"/>
+        <allow class="java.lang.reflect.Type"/>
+        <allow class="java.nio.charset.Charset"/>
```
rnveach: This shouldn't be needed since it is not imported in main code.
Please remove and confirm the others are needed.

src/main/java/com/github/checkstyle/regression/extract/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+/**
+ * Contains checkstyle maven command executing codes and extract file processor.
```
rnveach: `executing codes`?
Change to `Contains execution of checkstyle maven extract process and...`

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
```diff
@@ -0,0 +1,88 @@
...
+                .packageName(BASE_PACKAGE + ".checks")
+                .parent("Checker")
+                .build();
+        final String module2 = "EmptyStatementCheck";
+        final ModuleExtractInfo extractInfo2 = ImmutableModuleExtractInfo.builder()
+                .name(module2)
+                .packageName(BASE_PACKAGE + ".checks.coding")
+                .parent("TreeWalker")
+                .build();
+        assertEquals(2, map.size());
```
rnveach: We started requiring assertion messages in main repo by PMD.
This project doesn't have the newly updated PMD, but I may add it soon.

Please make sure all new assertions in PRs have message. It can be very simple phrase.
Check out main project for any examples.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        GSON = gsonBuilder.create();
+    }
+
+    /**
+     * Gets the full qualified name to module extract info map from the given reader.
+     * @param reader the given reader
+     * @return the full qualified name to module extract info map
+     */
+    public static Map<String, ModuleExtractInfo> getNameToModuleExtractInfoFromReader(
```
rnveach: `NameToModuleExtractInfo`
This makes it sound like we are passing to the method `name` as a parameter and it is giving us `ModuleExtractInfo`. We aren't giving it any parameters.
Please rename to something like `getModuleExtractInfoFromReader` or such.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            try {
+                modules = GSON.fromJson(reader, TYPE_EXTRACT_INFO_LIST);
+            }
+            finally {
+                reader.close();
+            }
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException(
```
rnveach: Throwing as a runtime exception is fine, but we will never be able to catch it in `Main` if we wish to add any more information to it. Please confirm this is your intent.

config/import-control.xml

src/main/java/com/github/checkstyle/regression/extract/package-info.java
Luolc: Done.

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
Luolc: The only way to go into catch block is `reader.close()` throwing an exception. I think it is rare. Instead of this, `GSON.fromJson` is more likely to throwing exception sometimes (wrong format, etc.). Json exception is not caught here so we could use them in main if needed.

rnveach: > Json exception is not caught here so we could use them in main 

Method has no `throws Exception` declaration so then all throwable exceptions, if any, are runtime.
We can't do anything about `GSON` but if we know an error is possible, we should not use runtime exceptions.
We can ignore this for now.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+    static {
+        final GsonBuilder gsonBuilder = new GsonBuilder();
+        for (TypeAdapterFactory factory : ServiceLoader.load(TypeAdapterFactory.class)) {
+            gsonBuilder.registerTypeAdapterFactory(factory);
+        }
+        GSON = gsonBuilder.create();
+    }
+
+    /**
+     * Gets the full qualified name to module extract info map from the given reader.
```
rnveach: Please remember to update Javadoc if you change method. This is the 3rd time now.
`Gets the full qualified name to module extract info map ` => `Gets the module extract info map`

Luolc: I did check this javadoc and I thought a "key to value map" description is clearer and kept it.

Luolc: My building is on a unexpected power cut god...
I can only use GitHub on my phone and there is no edit function on it...

Maybe `map of full qualified name to module extract info` is better? It refers the javadoc of the hardcode map in PackageObjectFactory. I am not able to write the link to it on my phone...

Luolc: I really think a key to value description is necessary. If  no description of the key in neither javadoc nor method name, it's confused what the key is when we use the method elsewhere

rnveach: @Luolc The issue is `name to module extract info` which is what method was named and we did away with it at https://github.com/checkstyle/regression-tool/pull/40#discussion_r124377270 .

rnveach: > a "key to value map" description is clearer 

How about this: `Gets the module extract info map from the given reader. Map key is the fully qualified module name.`

Luolc: It's good. :D Updated.

rnveach:
Merged.

@Luolc As you do only pieces of issue, please keep main issue updated with what is left to do in new post.

Issue #25, opened 2017-06-13
handle the tracking of deleted changes


Luolc:
Taken from https://github.com/checkstyle/regression-tool/pull/22#pullrequestreview-43713402

The `newPath` of deleted file is always `/dev/null` as mentioned at https://github.com/eclipse/jgit/blob/005e5feb4ecd08c4e4d141a38b9e7942accb3212/org.eclipse.jgit/src/org/eclipse/jgit/diff/DiffEntry.java#L373.
These could also reflected in our test https://github.com/Luolc/regression-tool/blob/c5d69e8a780fa84e7b414641f263e01e18a11f45/src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java#L87.

In the early discussion I want to keep "src/test" and deleted files mainly because I thought it is better to keep the possibility of getting all the changes, and leave a filter interface to let the real strategy code do decide which type of changes to use.

But I am re-thinking to drop deleted files now. If we would treat "ADD" and "MODIFY" the same way, then there is no need to distinguish the ChangeType. And "DELETE" is for sure useless as the file is gone.

rnveach:
> we would treat "ADD" and "MODIFY" the same way

I agree.

> "DELETE" is for sure useless as the file is gone.

The 2 important questions are what files are we tracking and what information can we grab from them on the specific change (add/delete/etc.).

For the type of files, we should be tracking modules, module tests, and utilities (anything not a module in production).
From modules, we need to know properties, messages, etc.
For module tests, we should be able to get some example property values.
For utilities, we need to work backwards from code that changed, to methods, to the modules involved if any.

If a module is deleted, there is nothing for us to test anymore. Module tests being deleted should be rare and there should be an accompanying additional tests. Utilities being deleted should be accompanied by changes in their respective modules or other utilities.

I don't see deletes helping us any. So I agree we should drop them in the `git` component.

rnveach:
This issue is blocked by PR #22 .

Luolc:
I am going to create a PR to drop DELETE changes in main code, then rebase this in #18 and update the DELETE related UTs.

rnveach:
@Luolc Can you just include the DELETE related UT in this PR then?
I specified this was blocked by the other PR. If you are going to do this first, than include a simple UT to showcase please.

Luolc:
[...]

PR #42, opened 2017-06-29
Issue #25: handle the tracking of deleted changes

Luolc:
#25

### DIFF
--- a/src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
+++ b/src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
@@ -50,6 +50,7 @@ public class DiffParserTest {
     @After
     public void tearDown() throws Exception {
         repository.close();
+        GitUtils.clearTempRepositories();
     }
 
     @Test
@@ -63,5 +64,18 @@ public class DiffParserTest {
                 repository.getDirectory().getParent(), "foo");
         assertEquals(1, changes.size());
         assertEquals("HelloWorld", changes.iterator().next().getPath());
+    }
+
+    @Test
+    public void testParseDeleteChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.addAnEmptyFileAndCommit(repository, "FileToDelete");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.removeFileAndCommit(repository, "FileToDelete");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be no changes detected", 0, changes.size());
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -64,4 +65,18 @@ public void testParse() throws Exception {
...
+    public void testParseDeleteChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            GitUtils.addAnEmptyFileAndCommit(repository, "HelloWorld");
+            GitUtils.addAnEmptyFileAndCommit(repository, "FileToDelete");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            GitUtils.removeFileAndCommit(repository, "FileToDelete");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be no changes detected", 0, changes.size());
+            FileUtils.deleteDirectory(repository.getDirectory().getParentFile());
```
rnveach: Shouldn't this go in `tearDown`?

Luolc: This `repository` is not the class field, it is the local variable created in this method. I would remove the `@Before` and class field way in #22

rnveach: You will have to think your test pattern than.
If anything throws an exception before it, delete will never happen and repository will be left on disk.
If you drop the field and don't have a before/after, than this should be in a `finally` which means you'll have to move `repository` into the correct scope.

Luolc:
```Java
    public static Repository createNewRepository() throws IOException {
        final File repoDir = File.createTempFile("TestTempRepository", "");
```
Since we are using `File.createTempFile` to create the directories, all of them would be in the default temp folder with the same prefix `TestTempRepository`.

Maybe the best way is: remove this line, add a `@After` and delete all the directories starting with `TestTempRepository`. Is that OK?

rnveach: > delete all the directories starting with TestTempRepository

It sounds more like a hack really.
Either have the finally or change `createNewRepository` to keep track of the `File`s and call a cleanup on in `After` on the list of files.

Luolc: Emm, I would create a repository pool to manage the directories and clean them all in `@After`.
I would @ you when I complete this.

Luolc: @rnveach Updated.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -64,4 +72,18 @@ public void testParse() throws Exception {
         assertEquals(1, changes.size());
         assertEquals("HelloWorld", changes.iterator().next().getPath());
     }
+
+    @Test
+    public void testParseDeleteChange() throws Exception {
+        try (Repository repository = GitUtils.createNewRepository()) {
+            repositories.add(repository);
```
rnveach: I am not expecting you to have to manually add the repository to the list.
`GitUtils.createNewRepository` should do this for you. You should also be calling a method in `GitUtils` to clear it.
The field should be private in it.

Luolc: Done.

Issue #20, opened 2017-06-11
Configuration Component


Luolc:
Taken from #2 
`configuration` should be given only the information it needs to create the configuration(s). This would be the list of modules and the specified properties.

rnveach:
@Luolc The following issue in Checkstyle must be completed for us to continue with 1 configuration with all modules: https://github.com/checkstyle/checkstyle/issues/4607

Luolc:
@rnveach OK. I am looking on that.

rnveach:
Issue is in our repo is done but we must finish issue in Checkstyle (https://github.com/checkstyle/checkstyle/issues/4607). I will leave this open to remind us of this.

rnveach:
Fix in checkstyle was merged.

PR #43, opened 2017-06-29
Issue #20: Configuration Component

Luolc:
#20 

Missing the output to file function currently.


config/pmd.xml
```diff
@@ -189,6 +189,8 @@
     <exclude name="MethodArgumentCouldBeFinal"/>
     <!--not configurable, decreases readability-->
     <exclude name="UseStringBufferForStringAppends"/>
+    <!--decreases readability when writing a muti-line string-->
+    <exclude name="AddEmptyString"/>
```
rnveach: XML strings should be in their own file, so this can be removed.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+    private static final String DOCTYPE_PUBLIC =
+            "-//Puppy Crawl//DTD Check Configuration 1.3//EN";
+
+    /** The "doctype-system" value of the config. */
+    private static final String DOCTYPE_SYSTEM =
+            "http://www.puppycrawl.com/dtds/configuration_1_3.dtd";
+
+    /** The config template. */
+    private static final String BASE_CONFIG = ""
+            + "<?xml version=\"1.0\" standalone=\"yes\"?>"
```
rnveach: This should be in it's own file and not hard coded into source.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+            + "<module name=\"Checker\">\n"
+            + "  <property name=\"charset\" value=\"UTF-8\"/>\n"
+            + "  <property name=\"severity\" value=\"warning\"/>\n"
+            + "  <property name=\"haltOnException\" value=\"false\"/>\n"
+            + "  <module name=\"BeforeExecutionExclusionFileFilter\">\n"
```
rnveach: Add comments founded in regression tool's XML.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml#L10

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Creates a new {@link Document} instance from the {@code BASE_CONFIG}.
+     * @return a new {@link Document} instance
+     */
+    private static Document createXmlDocument() {
+        final Document document;
+
+        try {
+            document = DocumentBuilderFactory.newInstance().newDocumentBuilder()
```
rnveach: This XML process shouldn't do any validating of DTD or request an internet connection to process.
Some users like to do regression solo when they are on a plane or such, so we need to limit internet connectivity when we can.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+                .getElementsByTagName(ELEMENT_MODULE);
+
+        for (int i = 0; i < nodeList.getLength(); ++i) {
+            final Node node = nodeList.item(i);
+            final Node nameAttr = node.getAttributes().getNamedItem(ATTR_NAME);
+            if (PARENT_TREE_WALKER.equals(nameAttr.getNodeValue())) {
+                returnValue = node;
+                break;
+            }
+        }
```
rnveach: empty line after `for`.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /** The attribute name "name". */
+    private static final String ATTR_NAME = "name";
+
+    /** The attribute name "value". */
+    private static final String ATTR_VALUE = "value";
+
+    /** The "doctype-public" value of the config. */
+    private static final String DOCTYPE_PUBLIC =
+            "-//Puppy Crawl//DTD Check Configuration 1.3//EN";
```
rnveach: Can we add a test to verify this doctype is latest version from Checkstyle release?
If this changes, I want us to be required to update this and verify it has no repercussions on regression.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+            + "<?xml version=\"1.0\" standalone=\"yes\"?>"
+            + "<module name=\"Checker\">\n"
+            + "  <property name=\"charset\" value=\"UTF-8\"/>\n"
+            + "  <property name=\"severity\" value=\"warning\"/>\n"
+            + "  <property name=\"haltOnException\" value=\"false\"/>\n"
+            + "  <module name=\"BeforeExecutionExclusionFileFilter\">\n"
+            + "    <property name=\"fileNamePattern\" value=\"module\\-info\\.java$\" />\n"
+            + "  </module>\n"
+            + "  <!-- Contents below are generated by regression-tool. -->\n"
+            + "  <module name=\"TreeWalker\"/></module>";
```
rnveach: The tool can't add `TreeWalker` for us?
If so, please explain the reason. Someday we may add other `TreeWalkers`. Just like we don't embed these modules into our extract, I would like to avoid embedding into other areas.

Luolc: > Someday we may add other TreeWalkers

I don't know whats the meaning of this, maybe smth like `MutiThreadTreeWalker` in the future?

I think by now, it is OK to let the tool generate `TreeWalker` dynamically. But there would be some other questions:
If we assume that the module would have whatever parent(not just `Checker` and `TreeWalker`), so how to decide where to put the parent module? Should we gain the info from the module extract info map? If yes, then configuration component would depend on module component, otherwise we would have to pass the whole info map to configuration component, but I don't think it is a good idea.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Do do mean that `document = DocumentBuilderFactory.newInstance().newDocumentBuilder()...` this line shouldn't do internet request? I switch off my internet connection and run the UTs of this class and everything is fine. Is that OK?

config/pmd.xml
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done.

src/test/java/com/github/checkstyle/regression/configuration/ConfigGeneratorTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Test
+    public void testIsProperUtilsClass() throws Exception {
+        assertUtilsClassHasPrivateConstructor(ConfigGenerator.class);
+    }
+
+    @Test
+    public void testGenerateConfigTextWithEmptyModuleInfos() throws Exception {
+        final String excepted = ""
+                + "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: > latest version from Checkstyle release

Is it OK to grab the latest version from this https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml? Is it required to be checkstyle release version but not SNAPSHOT version?

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: > I don't know whats the meaning of this, maybe smth like MutiThreadTreeWalker in the future?

TreeWalker = Java grammar.
If we add XML grammar, we will need something like XmlTreeWalker, etc.
MutiThreadTreeWalker is another possibility.

> configuration component would depend on module component

All components already rely on previous component. Next components input is previous component's output. I don't see any reason to embed the 2 together right now.

> how to decide where to put the parent module? Should we gain the info from the module extract info map?

Module extract info should already have this information ready for us to use.
[This is why we have parent and it is just String.](https://github.com/checkstyle/regression-tool/blob/master/src/main/java/com/github/checkstyle/regression/data/ModuleExtractInfo.java#L49)
Checker has no parent, so it is always root module. TreeWalker has parent Checker, so it goes as a sub-module of it. All checks have `TreeWalker` as parent, so they are sub-modules of it.
The extract can be looped through to understand hierarchy. If some hierarchy information is missing, we should throw an exception and say we can't continue.

> I think by now, it is OK to let the tool generate TreeWalker dynamically

But will it still require hardcoding of hierarchy somewhere? If so, let's move this to a new issue and keep code as is.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: That is fine. I just wanted you to be aware and verify. I am not familiar with the document builder.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: We can try this but internet connection may be an issue as travis uses anonymous user to access github.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: > But will it still require hardcoding of hierarchy somewhere?

Yep, by now we need to know the fact that "TreeWalker's parent is Checker", and that info should be gained through the module extract info map.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: So are we able to find the latest version of it from somewhere else?

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: Please create new issue then. We should avoid hardcoding things that can change in a single PR.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: without an internet connection, no. Github is the only place that has the latest information and it is stored in a separate repo.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,211 @@
...
+     * The information to create the node is grabbed from the given module info.
+     * @param document the XML document to create element
+     * @param moduleInfo the given module info
+     * @return a checkstyle module node
+     */
+    private static Node createModuleNode(Document document, ModuleInfo moduleInfo) {
+        final Element moduleNode = document.createElement(ELEMENT_MODULE);
+        moduleNode.setAttribute(ATTR_NAME, moduleInfo.name());
+
+        for (ModuleInfo.Property property : moduleInfo.properties()) {
```
Luolc: Done.

rnveach:
@Luolc appveyor should pass.

Luolc:
@rnveach Seems like a CRLF problem again, I am fixing it.

https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml#L4
https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml#L4

`http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd` and `http://www.puppycrawl.com/dtds/configuration_1_3.dtd`, which one should we use? They are different in checkstyle project and contribution repo.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Add a `testDtdVersion` in the test class. CI goods fine.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: > Please create new issue then. We should avoid hardcoding things that can change in a single PR.

#47 for this

rnveach:
> which one should we use?

We should always go by main project. We are forced to update and use things there.
Contribution is a 3rd party project that uses whatever they want. They aren't forced to use latest.

Luolc:
@rnveach I updated the url to `http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd`. CRLF is fixed and a test of dtd version is added.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,187 @@
...
+            }
+            else if (PARENT_TREE_WALKER.equals(parent)) {
+                treeWalkerNode.appendChild(moduleNode);
+            }
+        }
+
+        final Transformer transformer = createXmlTransformer();
+        final StringWriter writer = new StringWriter();
+        transformer.transform(new DOMSource(document), new StreamResult(writer));
+        return writer.getBuffer().toString().replaceAll("\r\n", "\n");
```
rnveach: > .replaceAll("\r\n", "\n");

Does this really need to be in production code? We are no longer doing any processing on this and Checkstyle/XML doesn't care about whitespace.
I assume this was added just to make tests pass, so why don't we keep it in test code.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,187 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+import com.github.checkstyle.regression.data.ModuleInfo;
+
+/**
+ * Generates the config XML and output it to a specific file.
```
rnveach: You return a String, not a file.
If this component is returning a file, the method needs to change. Otherwise this javadoc needs to change.

src/main/resources/com/github/checkstyle/regression/configuration/base_config.xml
```diff
@@ -0,0 +1,14 @@
...
+  <property name="severity" value="warning"/>
+  <!-- haltOnException is required for exception fixes and reporting of all exceptions -->
+  <property name="haltOnException" value="false"/>
+  <!-- BeforeExecutionFileFilters is required for sources of java9 -->
+  <module name="BeforeExecutionExclusionFileFilter">
+    <property name="fileNamePattern" value="module\-info\.java$"/>
+  </module>
+
+  <!-- Contents below are generated by regression-tool. -->
+  <module name="TreeWalker"/></module>
```
rnveach: Move `</module>` to next line.
You don't write code like:
````
void method() {
  if (condition) {} }
````

File: src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java (hunk 11 above)
Comment (Luolc) [inline src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java]: > Missing the output to file function currently.

I mentioned this at the first post of this PR. 

I am not sure which one(string or output to a file) is better. Personally I prefer output it to file, but I am also fine with a String. So I just left this when I first push this PR. What't your opinion.

File: src/main/resources/com/github/checkstyle/regression/configuration/base_config.xml (hunk 12 above)
Comment (Luolc) [inline src/main/resources/com/github/checkstyle/regression/configuration/base_config.xml]: I don't like this either, but there would be a strange indention problem if I put `</module>` to next line.

i.e.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
<module name="Checker">
  ...
  <!-- Contents below are generated by regression-tool. -->
  <module name="TreeWalker"/>
<module name="FileLengthCheck"/>
</module>
```

You could have a try to put that to next line, I assume `testGenerateConfigTextWithEmptyModuleInfos` and `testGenerateConfigTextWithCheckerParentModule` would fail due to that indention problem

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: Returning string to me seems weird as we will not be doing any more String manipulation and just output it to a file anyways. I would prefer it to return a file.

src/main/resources/com/github/checkstyle/regression/configuration/base_config.xml
rnveach: We shouldn't care about generated files as they are out of our control. I'm suprised that it even kept the comments.
Let's only concern ourselves with our files for now.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,187 @@
...
+     * @return the generated plain text of the config
+     * @throws TransformerException failure of transforming the XML document
+     */
+    public static String generateConfigText(List<ModuleInfo> moduleInfos)
+            throws TransformerException {
+        final Document document = createXmlDocument();
+
+        final Node checkerNode = getCheckerModuleNode(document);
+        final Node treeWalkerNode = getTreeWalkerModuleNode(document);
+        for (ModuleInfo moduleInfo : moduleInfos) {
```
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done.

src/main/resources/com/github/checkstyle/regression/configuration/base_config.xml
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done. `generateConfig` creates and output the xml plain text to a file. Related UTs are updated as well to directly compare whether two files are the same.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+        catch (TransformerConfigurationException ex) {
+            throw new IllegalStateException("cannot instantiate Transformer instance", ex);
+        }
+
+        return transformer;
+    }
+
+    /**
+     * Generates the config XML file from the given module inofs.
```
rnveach: `inofs` => `infos` I assume?

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
```diff
@@ -0,0 +1,207 @@
...
+     * @param moduleInfos the given module infos
+     * @return the generated config file
+     * @throws IOException          failure of creating and writing the file
+     * @throws TransformerException failure of transforming the XML document
+     */
+    public static File generateConfig(String path, List<ModuleInfo> moduleInfos)
+            throws IOException, TransformerException {
+        final File file = new File(path);
+        Files.write(file.toPath(), generateConfigText(moduleInfos)
+                .getBytes(Charset.forName("UTF-8")), StandardOpenOption.CREATE_NEW);
```
rnveach: Shouldn't we be using `TRUNCATE_EXISTING` instead of `CREATE_NEW`. We were told a path to place config, we should delete it if it already exists and not fail causing user to erase it manually.

src/test/java/com/github/checkstyle/regression/configuration/ConfigGeneratorTest.java
```diff
@@ -0,0 +1,164 @@
...
+        final ModuleInfo moduleInfo2 = ImmutableModuleInfo.builder()
+                .moduleExtractInfo(extractInfo2)
+                .build();
+
+        final File actual = generateConfig(Arrays.asList(moduleInfo1, moduleInfo2));
+        assertTrue("Config is not as expected", FileUtils.contentEquals(excepted, actual));
+    }
+
+    @After
+    public void tearDown() throws Exception {
```
rnveach: `DiffParserTest` had this at the top of the class while in here you put in the middle.
Let's be consistent and always have `After`/`Before` at the top of the class so it is the first thing users see when reading this file.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
Luolc: Done. I used `CREATE_NEW` for I was worried that user might accidentally write a path to another existing important file. But it is not common, I am fine with erasing.

src/test/java/com/github/checkstyle/regression/configuration/ConfigGeneratorTest.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java
rnveach: File name should be completely under our control. If user wants to save it, they should it move outside of our path. I don't expect users to have to supply a unique file path.

Luolc: Got it.

Luolc:
@rnveach Sorry I commited the changes but forgot to push. Updated are pushed now.

Issue #45, opened 2017-07-01
Fix diff problems when parsing a branch behind master


Luolc:
Taken from https://github.com/checkstyle/regression-tool/pull/22#issuecomment-307849787
If the PR branch hasn't rebased on the latest commits of master, the GitChange list would contain the files that changed in master but haven't been merged into PR branch yet. And these are not expected.

As mentioned at https://github.com/checkstyle/regression-tool/issues/18#issuecomment-312427445, `git diff master...<branch>` could work, but there is no directly implementation of it in JGit.
Now I find a synonymous command:
```bash
$ git diff $(git merge-base master <branch>)..<branch>
```
So we just need to implement `git merge-base master <branch>` and run the common diff command.
There is no directly `merge-base` command in JGit either, but it seems not hard to implement it by some workaround. Ref: https://stackoverflow.com/questions/26434452/how-is-a-merge-base-done-in-jgit

rnveach:
@Luolc Is there a command in JGit to find a common ancestor between 2 commits?
When a PR branch is behind master, it is still based on a specific commit in master, it is just not the latest commit.
If we find the common ancestor, we would only need to get the differences between that commit and the head PR commit.

Luolc:
I think there is no direct command in JGit to do that. In git CLI, it is `git merge-base`. But the `Git` class doesn't have a method `mergeBase`. But there could be some workaround to find the common ancestor, as I mentioned at the previous post.

rnveach:
Fix was merged

PR #46, opened 2017-07-02
Issue #45: Fix diff problems when parsing a branch behind master

Luolc:
#45 

Including refactor of `GitChange` since there starts to be `equals` use in the UTs.

### DIFF
--- a/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
+++ b/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
@@ -28,20 +28,20 @@ import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.diff.DiffEntry;
 import org.eclipse.jgit.lib.Constants;
-import org.eclipse.jgit.lib.ObjectReader;
-import org.eclipse.jgit.lib.Ref;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.RevCommit;
 import org.eclipse.jgit.revwalk.RevTree;
 import org.eclipse.jgit.revwalk.RevWalk;
+import org.eclipse.jgit.revwalk.filter.RevFilter;
 import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
 import org.eclipse.jgit.treewalk.AbstractTreeIterator;
 import org.eclipse.jgit.treewalk.CanonicalTreeParser;
 
 import com.github.checkstyle.regression.data.GitChange;
+import com.github.checkstyle.regression.data.ImmutableGitChange;
 
 /**
- * Parses git diff between feature branch and master for the further use.
+ * Parses git diff between PR branch and master for the further use.
  * @author LuoLiangchen
  */
 public final class DiffParser {
@@ -65,16 +65,13 @@ public final class DiffParser {
                 .readEnvironment().findGitDir().build();
 
         try {
+            final TreeParserPair pair = getTreeParserPair(repository, branchName);
             final Git git = new Git(repository);
 
             try {
-                final AbstractTreeIterator featureTreeParser =
-                        prepareTreeParser(repository, Constants.R_HEADS + branchName);
-                final AbstractTreeIterator masterTreeParser =
-                        prepareTreeParser(repository, Constants.R_HEADS + "master");
                 returnValue = git.diff()
-                        .setOldTree(masterTreeParser)
-                        .setNewTree(featureTreeParser)
+                        .setOldTree(pair.commonAncestorTreeParser)
+                        .setNewTree(pair.prTreeParser)
                         .call()
                         .stream()
                         .filter(entry -> entry.getChangeType() != DiffEntry.ChangeType.DELETE)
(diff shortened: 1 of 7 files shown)


src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -128,11 +120,55 @@ private static AbstractTreeIterator prepareTreeParser(Repository repository, Str
...
-        return new GitChange(diffEntry.getNewPath());
+        return ImmutableGitChange.builder()
+                .path(diffEntry.getNewPath())
+                .build();
+    }
+
+    /** A pair of tree parsers: feature tree parser and base tree parse. */
+    private static class TreeParserPair {
+        /** The tree parser of the head commit of feature branch. */
+        private AbstractTreeIterator featureTreeParser;
```
rnveach: Can we avoid using the words `feature` and `base` in this entire class? Can we stick to using master and PR branch naming?

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
```diff
@@ -65,6 +65,13 @@ public static void createNewBranchAndCheckout(Repository repository, String bran
         }
     }
 
+    public static void checkoutToBranch(Repository repository, String branchName)
```
rnveach: How about just `checkoutBranch` to follow the actual command?

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -93,32 +90,27 @@ private DiffParser() {
...
-
+            final RevCommit featureCommit = walk.parseCommit(
+                    repository.exactRef(Constants.R_HEADS + branch).getObjectId());
+            final RevCommit masterCommit = walk.parseCommit(
+                    repository.exactRef(Constants.R_HEADS + "master").getObjectId());
+            final RevCommit baseCommit = getBaseCommit(walk, featureCommit, masterCommit);
             walk.dispose();
+
+            returnValue.featureTreeParser = prepareTreeParser(walk, featureCommit);
+            returnValue.baseTreeParser = prepareTreeParser(walk, baseCommit);
```
rnveach: Shoudn't `TreeParserPair` have a constructor that takes these 2 as an argument to make sure they are always set together.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
Luolc: I am fine with `prBranch*`. 
But `base*` and `master*` have different meaning in my current code. I named the head commit of master as `master*`, and the common ancestor of PR branch and master as `base*`. 

"the common ancestor of PR branch and master" has a formal related command [merge-base](https://git-scm.com/docs/git-merge-base) in Git and I found others are used to calling the common commit as `baseCommit`. Thats why I use `base*`.

rnveach: Then please name them off master and such.

> common ancestor of PR branch and master as base*.

`masterCommitOfPr`?

Luolc: I don't think it is a good idea to name it with a `master*` prefix. master and PR branch have equal privilege in this context, `git merge-base master <pr-branch>` and `git merge-base <pr-branch> master` are exactly the same. A `master*` prefix is indicating that it is smth of master, but actually, it is not.

If `base*` is not good, I would prefer smth just like `common*`, `commonAncestor*`, or even `mergeBase*`.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
Luolc: Added a ctor now.

src/test/java/com/github/checkstyle/regression/internal/GitUtils.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
rnveach: `commonAncestor` is fine then.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -78,4 +79,23 @@ public void testParseDeleteChange() throws Exception {
             assertEquals("There should be no changes detected", 0, changes.size());
         }
     }
+
+    @Test
+    public void testParseFeatureBranchBehindMaster() throws Exception {
```
rnveach: `Feature` => `Pr`

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
Luolc: Done. I kept the function name `getMergeBase` in DiffParser, since it is just doing the related git command `git merge-base`. And the implementation and javadoc all refer to https://github.com/eclipse/jgit/blob/005e5feb4ecd08c4e4d141a38b9e7942accb3212/org.eclipse.jgit/src/org/eclipse/jgit/merge/Merger.java#L290

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -128,11 +120,67 @@ private static AbstractTreeIterator prepareTreeParser(Repository repository, Str
     }
 
     /**
+     * Gets the merge-base of two commits.
```
rnveach: > kept the function name getMergeBase in DiffParser, since it is just doing the related git command git merge-base. 

Please expand this documentation and explain what a merge-base is.

rnveach: In the terms we have talked about thus far.

Issue #21, opened 2017-06-11
Report Component


Luolc:
Taken from #2 
We will need report component, which runs the configuration(s) and pushes the reports to a server. We should try to re-use the groovy script in contribution repo.

Luolc:
> We should try to re-use the groovy script in contribution repo

I am thinking the way of "re-use". Via intuition, we could clone the contribution repo and run the groovy command line. Or, we could use the groovy source in some way, but it is a problem that how to maintain the groovy file to keep updating with the latest version.

@rnveach What't your suggestion?

rnveach:
To test the waters, we could just do a clone and see how it works out. Any 'issues' with this way can be made into a new issue so users are aware.

If there are no issues working with the clone, we could relocate the groovy scripts to this repo as we are both aiming for the same goal, regression testing. I see no reason right now why it shouldn't be in our repo instead.

We will face one problem with this component talked about at https://github.com/checkstyle/regression-tool/issues/5#issue-233729969 when we have multiple instances of the same module to test on.

Luolc:
I am meeting a lot of questions when working on this. I decide to push the changes step by step and confirm the unclear things in the meantime.

1. **listOfProjects**. The regression needs a lot of repositories of a of projects. It would be really time-consuming if we clone them every time we run it on Travis. And these repos tend to be consistent. Is it a good idea to add these repos in our regression-tool project directly?(I don't like this actually, but it seems the easiest way to avoid cloning every time) Or using some cache tech on Travis? I am not familiar with it and don't know whether it could work. There is another question that what projects to use for regression, I know the more the better, but time is a factor to be concerned about.
2. **diff.groovy**. Currently I wrote some code to download it from remote repo. Well, actually it is not so different with put it in our project directly. If we put it in our project, I think we may need a test to insure that it is the latest version comparing with the one on GitHub. There are many ways to run groovy in Java. By now the most easy-to-use way I found is `GroovyShell`. Do you have any suggestion on the way of running the groovy? FYI: http://groovy-lang.org/integrating.html
3. **configs**. Is it possible for us to use different configs for master and PR branch in regression-tool?
4. I think we may need git component to detect change type. Since we may use single mode when the module is new.

rnveach:
> It would be really time-consuming if we clone them every time we run it on Travis.
>  Or using some cache tech on Travis?

We have no choice on travis. We can't store everything into a cache and we are given new machines on each run.
[...]

PR #48, opened 2017-07-05
Issue #21: Report Component

Luolc:
#21 

> I decide to push the changes step by step and confirm the unclear things in the meantime.

This is not a completed PR currently.

As mentioned at https://github.com/checkstyle/regression-tool/issues/21#issuecomment-313109258, I wrote some code to download the script from remote repo. And actually it is not so different with putting it in our project directly. I assume we would only need `diff.groovy` in the contribution repo(and list of repos to test on), is it correct?

I think I need more discussion about this component. And here is the simple script downloading code, in case you have any suggestions about this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -315,6 +315,11 @@
                                         <branchRate>28</branchRate>
                                         <lineRate>52</lineRate>
                                     </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.report.ReportGenerator</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
+                                    </regex>
                                 </regexes>
                             </check>
                             <instrumentation>
--- a/src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
+++ b/src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
@@ -1,0 +1,69 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.report;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Generates the regression diff report.
+ * @author LuoLiangchen
+ */
+public final class ReportGenerator {
+    /** Prevents instantiation. */
+    private ReportGenerator() {
+    }
+
+    /**
+     * Generates the regression diff report.
+     * @param testerPath the path to the directory which holds 'diff.groovy'
+     * @param repoPath   the path to the checkstyle repository
+     * @param branch     the name of the PR branch
+     * @param configFile the generated config file
+     * @return the directory of the generated reports
+     * @throws InterruptedException failure of running CLI
+     * @throws IOException          failure of running CLI
+     */
+    public static File generate(
+            String testerPath, String repoPath, String branch, File configFile)
+            throws InterruptedException, IOException {
+        final Process process = new ProcessBuilder()
+                .directory(new File(testerPath))
+                .command(
+                        "groovy", "diff.groovy",
+                        "-r", repoPath,
+                        "-b", "master",
+                        "-p", branch,
+                        "-c", configFile.getAbsolutePath(),
+                        "-l", "projects-to-test-on.properties"
+                )
+                .inheritIO()
+                .start();
+        final int code = process.waitFor();
+        if (code != 0) {
+            throw new IllegalStateException("an error occurred when running diff.groovy");
+        }
+        final File reportDir = new File(testerPath, "reports/diff");
+        if (!reportDir.exists() || !reportDir.isDirectory()) {
+            throw new IOException("report does not exist or it is not a directory");
+        }
+        return reportDir;
+    }
+}
--- a/src/main/java/com/github/checkstyle/regression/report/package-info.java
+++ b/src/main/java/com/github/checkstyle/regression/report/package-info.java
@@ -1,0 +1,24 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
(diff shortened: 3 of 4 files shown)


rnveach:
>  I wrote some code to download the script from remote rep

I think we may need to take the contribution repo as an argument in Main.
The groovy scripts still call the contribution's `pom` so we can't just copy the scripts alone. The scripts also store repositories and such in relative directories and the output of the reports.
I originally wanted to move them here, but I think that may not be as easy.
For now, let's take the easy route and require it as an argument our program.
Sound good?

rnveach:
@Luolc ping

Luolc:
Invoking script with `GroovyShell` or smth else is impossible now. We could only use CLI way until we update `diff.groovy` and `launch.groovy`. CLI version is pushed.

We mentioned deployment should be in this component, and are we going to use the build-in deploy function in Travis currently? If yes, then this should be done by changes of `.travis.yml`

pom.xml
```diff
@@ -295,6 +295,11 @@
                                         <branchRate>28</branchRate>
                                         <lineRate>52</lineRate>
                                     </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.report.ReportGenerator</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
```
rnveach: Needs issue.

rnveach:
> I think we may need to take the contribution repo as an argument in Main.
> CLI version is pushed.

Please mention if you added something to issue like new argument. Since you edited, I didn't get a notification of the change.

> are we going to use the build-in deploy function in Travis currently? If yes, then this should be done by changes of .travis.yml

Please explain this as I am not sure what you mean by "built-in deploy".
If we want users to be able to deploy reports locally, they won't have travis built-in.

IMO, we should probably have different methods for deploy taken as arguments.
Ex: Github push, no deploy, etc...

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param testerPath the path to the directory which holds 'diff.groovy'
+     * @param repoPath   the path to the checkstyle repository
+     * @param branch     the name of the PR branch
+     * @param configPath the path to the generated config file
+     * @return the directory of the generated reports
+     * @throws InterruptedException failure of running CLI
+     * @throws IOException          failure of running CLI
+     */
+    public static File generate(
+            String testerPath, String repoPath, String branch, String configPath)
```
rnveach: https://github.com/checkstyle/regression-tool/blob/master/src/main/java/com/github/checkstyle/regression/configuration/ConfigGenerator.java#L126
Shouldn't `configPath` be a file?

I am unsure of other paths since we don't have code in `Main`. They should be converted to a `File` at some point to test if they exist and such.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+                        "groovy", "diff.groovy",
+                        "-r", repoPath,
+                        "-b", "master",
+                        "-p", branch,
+                        "-c", configPath,
+                        "-l", "projects-to-test-on.properties"
+                )
+                .inheritIO();
+        final Process process = builder.start();
+        final int code = process.waitFor();
```
rnveach: Will users' see output of groovy command?

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+        if (code != 0) {
+            throw new IllegalStateException("an error occurs when running diff.groovy");
```
rnveach: `occurs` => `occurred`

config/pmd.xml
```diff
@@ -64,6 +64,12 @@
     <!-- Checkstyle is not thread safe -->
     <exclude name="UseConcurrentHashMap"/>
   </rule>
+  <rule ref="rulesets/java/controversial.xml/UseObjectForClearerAPI">
+    <properties>
+      <!-- Considering fix this later, since the interface might be changed. -->
+      <property name="violationSuppressXPath" value="//ClassOrInterfaceDeclaration[@Image='ReportGenerator']"/>
```
rnveach: Make issue.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
Luolc: I am OK with a `File` parameter. Is there any possibility that user might use his own config file and don't use the generated one?

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
Luolc: Yes. `inheritIO()` makes users see output of the CLI. We could remove `inheritIO()` if we don't want user see the output.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
rnveach: > Is there any possibility that user might use his own config file and don't use the generated one?

It is a possibility, but like I stated, anything that comes from user has to converted to a `File` at some point to test if it exists and such. We should probably do this sooner rather than later.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
rnveach: They need to see output as any error will be in groovy and not displayed by our process.
This is why I wanted to confirm.

pom.xml
Luolc: #55

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
Luolc: Done.

config/pmd.xml
Luolc: This is removed after we changed parameter `configPath` to a `File` type.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
Luolc: Updated. It is now a `File` type. And I add a check of the existence. Do you think it is proper to check here, or in `Main`, before passing it to the method?

rnveach: `Main` should do all validation to check if inputs are correct.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param configFile the generated config file
+     * @return the directory of the generated reports
+     * @throws InterruptedException failure of running CLI
+     * @throws IOException          failure of running CLI
+     */
+    public static File generate(
+            String testerPath, String repoPath, String branch, File configFile)
+            throws InterruptedException, IOException {
+        if (!configFile.exists()) {
+            throw new IllegalArgumentException("config file does not exist: " + configFile);
```
rnveach: Like I said, `Main` should do all validation and testing to check if inputs are valid.
All components should assume it is given valid information.

Please remove this `if` check.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
```diff
@@ -0,0 +1,68 @@
...
+                        "-c", configFile.getAbsolutePath(),
+                        "-l", "projects-to-test-on.properties"
+                )
+                .inheritIO()
+                .start();
+        final int code = process.waitFor();
+        if (code != 0) {
+            throw new IllegalStateException("an error occurred when running diff.groovy");
+        }
+        return new File(testerPath, "reports/diff");
```
rnveach: Should we test if it exists and is directory before continuing? We don't want to return bad data.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
```diff
@@ -0,0 +1,68 @@
...
+    public static File generate(
+            String testerPath, String repoPath, String branch, File configFile)
+            throws InterruptedException, IOException {
+        if (!configFile.exists()) {
+            throw new IllegalArgumentException("config file does not exist: " + configFile);
+        }
+        final Process process = new ProcessBuilder()
+                .directory(new File(testerPath))
+                .command(
+                        "groovy", "diff.groovy",
```
rnveach: One final thing, let's make a new issue on using GroovyShell.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/report/ReportGenerator.java
Luolc: #56

PR #50, opened 2017-07-06
Issue #37: inject, invoke maven command, read file and clean up

Luolc:
#37 

This PR is just showing the abstract design thinking as mentioned at https://github.com/checkstyle/regression-tool/issues/37#issuecomment-313466371

We need do the following things:
1. run Git command to checkout to the PR branch.
2. move the extract program to checkstyle repo
3. invoke maven command
4. read the file to memory
5. clear

Problems:
1. Where to put step 1? According to our design of package, there would be no JGit dependency outside `git`. So should we create a util class in `git` and use it in `extract`? If yes, we already have a `GitUtils` in test scope, we might need to rename it to smth like `GitTestUtils` and create a `GitUtils` for production code. If confirmed, we should make split issue for it and discuss there.
2. `CheckstyleInjector` is a temporary name. I haven't thought out a proper name currently. Please share your suggestion if you have good ideas of the name.
3. I would like to keep `getModuleExtractInfosFromReader` in `ExtractInfoProcessor` public, for we need this middleware method to create fake infos in `ModuleInfoCollectorTest` and `ModuleUtilsTest`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,11 @@
             <version>2.5</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.maven.shared</groupId>
+            <artifactId>maven-invoker</artifactId>
+            <version>3.0.0</version>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
             <version>2.8.1</version>
@@ -288,7 +293,22 @@
                                     <regex>
                                         <pattern>com.github.checkstyle.regression.configuration.ConfigGenerator</pattern>
                                         <branchRate>80</branchRate>
-                                        <lineRate>91</lineRate>
+                                        <lineRate>92</lineRate>
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.ExtractInfoProcessor</pattern>
+                                        <branchRate>100</branchRate>
+                                        <lineRate>60</lineRate>
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.CheckstyleInjector.*</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.InjectException</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
                                     </regex>
                                     <regex>
                                         <pattern>com.github.checkstyle.regression.module.ModuleUtils</pattern>
--- a/src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
+++ b/src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
@@ -1,0 +1,168 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.extract;
+
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.maven.shared.invoker.DefaultInvocationRequest;
+import org.apache.maven.shared.invoker.DefaultInvoker;
+import org.apache.maven.shared.invoker.InvocationRequest;
+import org.apache.maven.shared.invoker.InvocationResult;
+import org.apache.maven.shared.invoker.Invoker;
+import org.apache.maven.shared.invoker.MavenInvocationException;
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+
+/**
+ * Injects files to checkstyle repository, which would be invoked by maven command
+ * to generate module extract info.
+ * @author LuoLiangchen
+ */
+final class CheckstyleInjector implements Closeable {
+    /** The path to checkstyle repository. */
+    private final String repoPath;
+
+    /** The name of PR branch. */
+    private final String branch;
+
+    /** The checkstyle repository. */
+    private final Repository repository;
+
+    /**
+     * Creates a new instance of CheckstyleInjector.
+     * @param repoPath the path to checkstyle repository
+     * @param branch   the name of PR branch
+     */
+    CheckstyleInjector(String repoPath, String branch) {
+        this.repoPath = repoPath;
+        this.branch = branch;
+
+        final File gitDir = new File(repoPath, ".git");
+        final Repository repo;
+        try {
+            repo = new FileRepositoryBuilder().setGitDir(gitDir)
+                    .readEnvironment().findGitDir().build();
+        }
+        catch (IOException ex) {
+            throw new IllegalStateException(ex);
+        }
+        repository = repo;
+    }
+
+    /**
+     * Generates the module extract info file.
+     * @return the module extract info file
+     * @throws InjectException failure of generation
+     */
+    public File generateExtractInfoFile() throws InjectException {
+        try {
+            checkoutToPrBranch();
+            copyInjectFilesToCheckstyleRepo();
+            invokeMavenCommand();
+            return new File(repoPath, "checkstyle_modules.json");
+        }
+        catch (IOException | GitAPIException ex) {
+            throw new InjectException("unable to generate extract info file", ex);
+        }
+    }
+
+    /**
+     * Clears the injected files and the generated info file in checkstyle repository.
+     * @throws InjectException failure of clearing
+     */
+    public void clearInjections() throws InjectException {
+        final Git git = new Git(repository);
+
+        try {
+            git.clean().setCleanDirectories(true).call();
+        }
+        catch (GitAPIException ex) {
+            throw new InjectException("unable to clear injections", ex);
+        }
+        finally {
+            git.close();
+        }
+    }
+
+    /** Closes the repository resource. */
+    @Override
+    public void close() {
+        repository.close();
+    }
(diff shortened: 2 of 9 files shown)


rnveach:
> 1. Where to put step 1? 
> there would be no JGit dependency outside git. 

truthfully I was thinking we would run a command line git execution instead of using JGit.
I like the idea of using JGit to make it truly system independent and not relying on something being installed.
If we are doing this for JGit, it is possible to do this for Maven too. See [maven-embedder](https://github.com/checkstyle/contribution/compare/master...rnveach:regression#diff-f7409c8bfb778bdc0abb91ba3ec2b912R21).

> So should we create a util class in git and use it in extract? 

I didn't want to allow jgit everywhere as then someone will import it in the wrong place and think it is acceptable when it isn't. Import control is to restrict bad designs, not to limit our design possibilities.
I don't think one component should be calling another, which is what a utility class in `git` will look like.
We can either allow jgit inside `extract` component (I think I prefer this right now), or make a `util` package and name class something like `GitUtil`.

> 2. CheckstyleInjector 

`CheckstyleExtraction`. Our main purpose to extract information to checkstyle after all.

> 3. I would like to keep getModuleExtractInfosFromReader in ExtractInfoProcessor public, for we need this middleware method to create fake infos in ModuleInfoCollectorTest and ModuleUtilsTest.

No. If something is needed for testing, we should use reflection and break class that way than leave internal structure open for someone to mis-use. Production should not be defined by tests except for maybe splitting things into smaller pieces.
In my work, if I need a method like this, I use reflection and store it as a class field for use everywhere in test.

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @author LuoLiangchen
+ */
+final class CheckstyleInjector {
+    static File generateExtractInfoFile(String repoPath, String branch)
+            throws IOException, GitAPIException {
+        checkoutToPrBranch(repoPath, branch);
+        moveInjectFilesToCheckstyleRepo(repoPath);
+        invokeMavenCommand(repoPath);
+        return null;
```
rnveach: missing `readExtractFile();` :)

Luolc:
> No. If something is needed for testing, we should use reflection and break class that way than leave internal structure open for someone to mis-use.

OK.

> We can either allow jgit inside extract component (I think I prefer this right now)

I agree with this.

> it is possible to do this for Maven too. See maven-embedder

Ya, I was thinking to use it. Already been investigating on it.

Luolc:
@rnveach Updated. 
missing implementation of `invokeMavenCommand` currently. 

I don't use static method in `CheckstyleInjector` this time, for every method in it needs `repoPath` and `branch`, and we need close `repository`, I think an instance is better.

Maybe we need to combine the another part of this issue then we could make test for all this updates.

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
```diff
@@ -0,0 +1,140 @@
...
+     * Generates the module extract info file.
+     * @return the module extract info file
+     * @throws IOException     failure of generation
+     * @throws GitAPIException JGit library exception
+     */
+    public File generateExtractInfoFile() throws IOException, GitAPIException {
+        checkoutToPrBranch();
+        copyInjectFilesToCheckstyleRepo();
+        invokeMavenCommand();
+        return null;
```
rnveach: Needs a result.

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public void close() {
+        repository.close();
+    }
+
+    /**
+     * Invokes Maven command to generate the extract info file in checkstyle repository.
+     */
+    private static void invokeMavenCommand() {
+        // invoke maven command to generate the extract info file
```
rnveach: needs to do something.

pom.xml
```diff
@@ -288,7 +288,17 @@
...
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.ExtractInfoProcessor</pattern>
+                                        <branchRate>100</branchRate>
+                                        <lineRate>66</lineRate>
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.CheckstyleInjector</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
```
rnveach: Please create issue to create some tests and coverage for this class at a later time.

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
Luolc: Updated. But have some problems. See my comments below.

Luolc:
@rnveach I encountered some strange problems:
1. cobertura failed. I add/update the regex pattern according to the html report, but it still failed.
2. If I uncomment the UT added in `ExtractInfoProcessorTest`: https://github.com/Luolc/regression-tool/blob/689f22c5d44223c268fe23227b8311bdae039e26/src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java#L105, `mvn clean test` is fine but `mvn clean compile cobertura:cobertura` would run with a lot of test error. 
If comment/remove this UT, then there would be no error with `compile cobertura`. I guess it is caused by `System.setProperty(PROPERTY_KEY_DIRECTORY, repoPath);`, and this is because the problem described in https://stackoverflow.com/questions/33400574/unable-to-run-maven-tasks-through-mavencli-maven-embedder. Can't fix this after a long time of investigation. :(

rnveach:
> cobertura failed

remove `quiet` from POM for cobertura and you get this error:
> [ERROR] com.github.checkstyle.regression.extract.CheckstyleInjector$1 failed coverage check. Line coverage rate of 0.0% is below 100.0%

rnveach:
> If I uncomment the UT added in ExtractInfoProcessorTest

This is why I have said repeatedly, lets not worry about code coverage on cloning, maven, etc...
Isolate it as best as you can and mock its true run away. We will eventually use travis to make sure it functions properly in a full run.

>  guess it is caused by System.setProperty(PROPERTY_KEY_DIRECTORY, repoPath);

Why not set this as arguments to `doMain` as specified in https://stackoverflow.com/questions/33400574/unable-to-run-maven-tasks-through-mavencli-maven-embedder
Ex: `int result = cli.doMain(new String[] { "-Dmaven.multiModuleProjectDirectory=" + projectRoot, "install" }, "/path/to/project", System.out, System.err);`

Luolc:
> We will eventually use travis to make sure it functions properly in a full run.

I add this in UT just for quick test and debug, I would remove it when I confirm the code function is acceptable.

> Why not set this as arguments to doMain as specified in 

I tried this but it doesn't work. Still has error `-Dmaven.multiModuleProjectDirectory system property is not set. Check $M2_HOME environment variable and mvn script match.`

Luolc:
> Change line to something like: <pattern>com.github.checkstyle.regression.extract.CheckstyleInjector.*</pattern>

It works. Thanks. :)

rnveach:
https://github.com/checkstyle/regression-tool/pull/50#discussion_r126655932
This is not done.

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
```diff
@@ -72,21 +84,35 @@ public void testGetModuleExtractInfosFromReader() {
...
+        catch (InvocationTargetException ex) {
             assertEquals(
                     "Exception message is wrong",
                     "Failed when loaing hardcoded checkstyle module information",
-                    ex.getMessage());
+                    ex.getTargetException().getMessage());
         }
     }
+
+    //@Test
```
rnveach: Do not leave commented code in.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -58,12 +62,45 @@ private ExtractInfoProcessor() {
...
+
+        try {
+            try {
+                final File file = injector.generateExtractInfoFile();
+                final Reader reader = new InputStreamReader(
+                        new FileInputStream(file), Charset.forName("UTF-8"));
+                returnValue = getModuleExtractInfosFromReader(reader);
+            }
+            catch (IOException ex) {
+                throw new InjectException("unable to read to info file", ex);
```
rnveach: This message is not very informative of issue.
Expand on what we were trying to do. Since it is the extract file, it should be in message somewhere.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
```diff
@@ -58,12 +62,45 @@ private ExtractInfoProcessor() {
...
+     * @throws InjectException failure when making injection
+     */
+    public static Map<String, ModuleExtractInfo> getModuleExtractInfos(
+            String repoPath, String branch) throws InjectException {
+        final Map<String, ModuleExtractInfo> returnValue;
+        final CheckstyleInjector injector = new CheckstyleInjector(repoPath, branch);
+
+        try {
+            try {
+                final File file = injector.generateExtractInfoFile();
```
rnveach: if `generateExtractInfoFile` doesn't throw an `IOException` it should be outside the 2nd level `try`.

pom.xml
```diff
@@ -288,7 +293,22 @@
...
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.CheckstyleInjector.*</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.InjectException</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
```
rnveach: Needs issue.

pom.xml
```diff
@@ -288,7 +293,22 @@
...
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.ExtractInfoProcessor</pattern>
+                                        <branchRate>100</branchRate>
+                                        <lineRate>60</lineRate>
+                                    </regex>
+                                    <regex>
+                                        <pattern>com.github.checkstyle.regression.extract.CheckstyleInjector.*</pattern>
+                                        <branchRate>0</branchRate>
+                                        <lineRate>0</lineRate>
```
rnveach: Needs issue.

src/test/java/com/github/checkstyle/regression/extract/ExtractInfoProcessorTest.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/extract/ExtractInfoProcessor.java
Luolc: Updated.

pom.xml
Luolc: #54

pom.xml
Luolc: #54

PR #51, opened 2017-07-06
Pull #51: fix write option bug in ConfigGenerator

Luolc:
`TRUNCATE_EXISTING` option is only for there is already a file exiting.
`CREATE` is necessary for nothing there, without is there would be exception raised.
These two could be set together.


rnveach:
@Luolc This is not minor as it is a bug fix. Please reference this PR in commit message.
Please add a test to showcase problem.

Luolc:
So need I create an issue for this?

rnveach:
No issue is needed, referencing PR in commit message is enough.
PR has basically become the issue.

Luolc:
@rnveach Done.

PR #52, opened 2017-07-06
Issue #37: extract program

Luolc:
#37

This is the file that would be injected into checkstyle project.

Also we should update `pom.xml` in checkstyle to add a Gson dependency.

- [x] name
- [x] packageName
- [x] parent
- [x] interfaces
- [x] hierarchies
- [ ] properties
- [ ] tokens

As mentioned at https://github.com/checkstyle/regression-tool/pulls#issuecomment-310935967, we ignored the last two for now.

**Edit**: If necessary, I am fine with combining this PR with #50 into one PR together.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,20 @@ matrix:
         - DESC="findbugs and pmd"
         - CMD="mvn clean compile pmd:check findbugs:check"
 
+    # checkstyle injection compile no error (oraclejdk8)
+    - jdk: oraclejdk8
+      env:
+        - DESC="checkstyle injection compile no error"
+        - CMD1="git clone https://github.com/checkstyle/checkstyle &&"
+        - CMD2="cp src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java "
+        - CMD3="checkstyle/src/test/java/com/puppycrawl/tools/checkstyle/ExtractInfoGeneratorTest.java &&"
+        - CMD4="cp src/main/resources/com/github/checkstyle/regression/extract/JsonUtil.java "
+        - CMD5="checkstyle/src/test/java/com/puppycrawl/tools/checkstyle/JsonUtil.java &&"
+        - CMD6="cd checkstyle &&"
+        - CMD7="mvn clean compile &&"
+        - CMD8="mvn test -Dtest=ExtractInfoGeneratorTest#generateExtractInfoFile"
+        - CMD="$CMD1$CMD2$CMD3$CMD4$CMD5$CMD6$CMD7$CMD8"
+
 script: eval $CMD
 
 after_success:
--- a/src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
+++ b/src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
@@ -1,0 +1,80 @@
+package com.puppycrawl.tools.checkstyle;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+
+import org.junit.Test;
+
+import com.puppycrawl.tools.checkstyle.internal.CheckUtil;
+import com.puppycrawl.tools.checkstyle.utils.ModuleReflectionUtils;
+
+/**
+ * This file would be injected into checkstyle project and invoked by maven command
+ * to generate the module extract info file.
+ * @author LuoLiangchen
+ */
+public class ExtractInfoGeneratorTest {
+    /**
+     * Generates the extract info file named as "checkstyle_modules.json".
+     * @throws IOException failure when generating the file
+     */
+    @Test
+    public void generateExtractInfoFile() throws IOException {
+        final List<Class<?>> modules = new ArrayList<>(CheckUtil.getCheckstyleModules());
+        modules.sort(Comparator.comparing(Class::getSimpleName));
+        final JsonUtil.JsonArray moduleJsonArray = new JsonUtil.JsonArray();
+        for (Class<?> module : modules) {
+            moduleJsonArray.add(createJsonObjectFromModuleClass(module));
+        }
+        String jsonString = moduleJsonArray.toString();
+        final File output = new File("checkstyle_modules.json");
+        Files.write(output.toPath(), jsonString.getBytes(Charset.forName("UTF-8")),
+                StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
+    }
+
+    /**
+     * Creates Json object for a module from the module class.
+     * @param clazz the given module class
+     * @return the Json object describing the extract info of the module
+     */
+    private static JsonUtil.JsonObject createJsonObjectFromModuleClass(Class<?> clazz) {
+        final JsonUtil.JsonObject object = new JsonUtil.JsonObject();
+
+        final String name = clazz.getSimpleName();
+        final String parent;
+        if (ModuleReflectionUtils.isCheckstyleCheck(clazz)) {
+            parent = "TreeWalker";
+        }
+        else if (ModuleReflectionUtils.isRootModule(clazz)) {
+            parent = "";
+        }
+        else {
+            parent = "Checker";
+        }
+        object.addProperty("name", name);
+        object.addProperty("packageName", clazz.getPackage().getName());
+        object.addProperty("parent", parent);
+
+        final JsonUtil.JsonArray interfaces = new JsonUtil.JsonArray();
+        final JsonUtil.JsonArray hierarchies = new JsonUtil.JsonArray();
+        Arrays.stream(clazz.getInterfaces()).forEach(cls -> interfaces.add(cls.getCanonicalName()));
+        Class<?> superClass = clazz.getSuperclass();
+        while (!Object.class.equals(superClass)) {
+            hierarchies.add(superClass.getCanonicalName());
+            Arrays.stream(superClass.getInterfaces())
+                    .forEach(cls -> interfaces.add(cls.getCanonicalName()));
+            superClass = superClass.getSuperclass();
+        }
+        object.add("interfaces", interfaces);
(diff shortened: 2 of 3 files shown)


rnveach:
> pom.xml in checkstyle should be added a Gson dependency.

Checkstyle doesn't want any changes in their repository for our process, so this won't be possible.

See my comment at https://github.com/checkstyle/regression-tool/issues/37#issue-238414241 :
> So we don't have to inject gson dependency into Checkstyle too, I think we should just write out file by hand.

Otherwise, we have to inject test dependency into their POM, which may not be pretty and require us undoing the change after our process.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.Test;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
```
rnveach: See my comments.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class ExtractInfoGeneratorTest {
+    /**
+     * Generates the extract info file named as "checkstyle_modules.json".
+     * @throws IOException failure when generating the file
+     */
+    @Test
+    public void generateExtractInfoFile() throws IOException {
+        final Gson gson = new GsonBuilder().setPrettyPrinting().create();
+        final List<JsonObject> objects = CheckUtil.getCheckstyleModules().stream()
```
rnveach: One issue with `CheckUtil.getCheckstyleModules` is we can't test on old commits.
I think we should find someway to be more backward compatible. This can be a new issue.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    private static JsonObject createJsonObjectFromModuleClass(Class<?> clazz) {
+        final JsonObject object = new JsonObject();
+
+        final String name = clazz.getSimpleName();
+        final String parent;
+        if (ModuleReflectionUtils.isCheckstyleCheck(clazz)) {
+            parent = "TreeWalker";
+        }
+        else if ("Checker".equals(name)) {
```
rnveach: Check if it is a root module instead of `Checker`.

Luolc:
> Checkstyle doesn't want any changes in their repository for our process, so this won't be possible.
> Otherwise, we have to inject test dependency into their POM, which may not be pretty and require us undoing the change after our process.

I don't mean that we need I change in checkstyle main repo. At the meantime we do the injection, we just need to find this line, https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L267, and add the dependency below the line. If `pom.xml` changes one day, the maven process would fail quickly and then we would fix it. And I think this way is much better than we generate the file by our hand.

All the clear up operation can be done with a simple `git checkout .`, whatever JGit or just running command line Git. So we don't need to be worried about clearing.

rnveach:
> If pom.xml changes one day, the maven process would fail quickly and then we would fix it. 

Whereas if we write file by hand, it will never fail.
What are we gaining by adding this dependency besides easier to make/maintain code?
Adding dependency will make it harder to test in a travis build as travis was going to be a separate process from our production code.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
Luolc: Done.

Luolc:
@rnveach Updated with our own json generation codes.

rnveach:
> We need a way to ensure this file is compilable, especially in checkstyle process.

@Luolc Can you update travis to make sure this code compiles and executes without exception?

Luolc:
@rnveach Done.

.travis.yml
```diff
@@ -46,6 +46,19 @@ matrix:
...
+      env:
+        - DESC="checkstyle injection compile no error"
+        - CMD1="git clone https://github.com/checkstyle/checkstyle &&"
+        - CMD2="cp src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java "
+        - CMD3="checkstyle/src/test/java/com/puppycrawl/tools/checkstyle/ExtractInfoGeneratorTest.java &&"
+        - CMD4="cp src/main/resources/com/github/checkstyle/regression/extract/Json.java "
+        - CMD5="checkstyle/src/test/java/com/puppycrawl/tools/checkstyle/Json.java &&"
+        - CMD6="cd checkstyle &&"
+        - CMD7="mvn clean compile"
+        - CMD="$CMD1$CMD2$CMD3$CMD4$CMD5$CMD6$CMD7"
```
rnveach: > We can go further and test execution of file to make sure it doesn't atleast throw exception or fail.

What about this statement?

src/main/resources/com/github/checkstyle/regression/extract/Json.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Our own Json processing classes.
+ * @author LuoLiangchen
+ */
+class Json {
```
rnveach: `JsonUtil`

src/main/resources/com/github/checkstyle/regression/extract/Json.java
```diff
@@ -0,0 +1,114 @@
+package com.puppycrawl.tools.checkstyle;
+
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Our own Json processing classes.
```
rnveach: `processing classes` => `processing utility`.

.travis.yml
Luolc: Maybe better in #50 ? The test on generated file needs a lot of process of read/convert/... And many logic are included in #50 . Should we combine the changes of these two PR and close one of them?

src/main/resources/com/github/checkstyle/regression/extract/Json.java
Luolc: Done.

src/main/resources/com/github/checkstyle/regression/extract/Json.java
Luolc: Done.

.travis.yml
rnveach: I am fine with this being separated.
I don't expect this to be done in #50 in the junits (if that was what your thinking), as junits won't be cloning the external repository as part of their tests. They should be pretty quick tests, which is why missing code coverage for this one script is fine. That is why travis makes better sense for doing all these long complicated tasks.

Once you add this to travis I will merge this as it will be confirmed working.

rnveach:
@Luolc One final thing, can you upload the result of this process somewhere for us to look at during GSoC?

.travis.yml
Luolc: What level of the test should be? The ones I could think out now:
1. check it is a valid json
2. check whether it is in correct format(a list, each item contains "name", "packageName" etc)
Do you have anything to add?

And, to test a json format is much hard to do with only shell script, if we won't do it in UT, then could we use other script lang like Groovy?

Luolc:
@rnveach Do you mean upload it every time sending a PR? I may try to add a deploy operation in Travis.

.travis.yml
rnveach: @Luolc no exception test is fine for now.
Travis won't be verifying output for now. We can manually confirm this.

Later since this extract program should work on all checkstyle versions, we could use release for code coverage and expected output in another issue.

rnveach:
@Luolc I meant a one time upload in a place where we can look back on it in the future.
Every PR is not needed.

Luolc:
@rnveach I upload a gist: https://gist.github.com/Luolc/f3898dc507d66d21b48d85c1c09e5001

.travis.yml
Luolc: @rnveach Don't exactly understand the way actually...
But let's do the further discussion in the issue for this, if you think current test is fine for now. :) Should I create that issue now?

rnveach:
> Don't exactly understand the way actually...

If you mena the way to execute the test to make sure it passes with no errors, see https://github.com/checkstyle/regression-tool/issues/37#issuecomment-313125292
> Execute extract program as a one time junit. See https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/64d616821b2d2265fb8291fd6950b4eef83c4547/.travis.yml#L68 on how to do this execution.

Luolc:
@rnveach Oh I think I misunderstood this. I was thinking to test the json format. If you mean run this UT, I am clear of the way.

rnveach:
I assume you already tested the json format since you had it originally in our repo with the full thing but than trimmed it down for just what we used in tests.
What is in gist can also be tested by itself.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    private static JsonUtil.JsonObject createJsonObjectFromModuleClass(Class<?> clazz) {
+        final JsonUtil.JsonObject object = new JsonUtil.JsonObject();
+
+        final String name = clazz.getSimpleName();
+        final String parent;
+        if (ModuleReflectionUtils.isCheckstyleCheck(clazz)) {
+            parent = "TreeWalker";
+        }
+        else if (RootModule.class.isAssignableFrom(clazz)) {
```
rnveach: `RootModule.class.isAssignableFrom` => `ModuleReflectionUtils.isRootModule`

Luolc: Done.

Issue #32, opened 2017-06-16
Main


rnveach:
We need a main class that users will use to access and run 'regression-tool'.

It needs to take in custom parameters, validate the parameters, and use those values to execute the tool.
If any parameters are wrongly entered, we need to give an informative error message and exit with non-zero code.
If an exception happens in any component, we should rethrow it all the way up to the Main class and display it to the user.

What kind of parameters will we accept? Which are required, which can be optional, and what will be the default values of the optional ones?
We will need a type of help menu, to display to users the list of parameters and their brief descriptions and anything else they need to know to use them correctly.

Blocked by Issue #19, #20, #21 .

Luolc:
Arguments we found by now:
- **repository path**: the path of the repository. Required.
- **branch name**: the name of the PR branch. Required.

More arguments are need in `report` component. Those would depend on the discussion in #21

rnveach:
@Luolc Once we are finished with all components up to and including generating config, we need to start this.
@romani wants a simple test program that will generate configs for users.

Other components may take a back seat and we may start working on expanding what is getting placed into the generated config file.

rnveach:
@Luolc Please start this.
Config generation only should be some type of special option. It should not be the default.

Luolc:
@rnveach 
> Config generation only should be some type of special option.

Do you mean that we have two options: 1. generate config and report 2. only generate config ?

rnveach:
> Do you mean that we have two options: 1. generate config and report 

1 should be the default as this will be the most common case when our utility is fully functional. Users shouldn't need to give an option to do this, it should be default.
Not doing report should be some kind of a special option like, `-dontGenerateReport` or `-stopAfterGenerateConfig`.
We are not a configuration generation tool, we are a full regression tool. So our default behavior should be just that.

rnveach:
Issues have been created for remaining work, so I am closing this

PR #57, opened 2017-07-20
Issue #32: Main

Luolc:
I don't know what to add in `pom.xml`. I failed to run the project by `java -jar ...` way.

I haven't handle any exceptions of components now, how should be present the exception message for user? Should we have a `[ERROR]` prefix or smth like this?

### DIFF
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -4,6 +4,9 @@
         "http://www.puppycrawl.com/dtds/import_control_1_1.dtd">
 
 <import-control pkg="com.github.checkstyle.regression">
+    <allow pkg="org.apache.commons.cli" local-only="true"/>
+    <allow pkg="com.github.checkstyle.regression" local-only="true"/>
+    <allow pkg="java.text" local-only="true"/>
     <allow pkg="java.io"/>
     <allow pkg="java.nio"/>
     <allow pkg="org.apache.commons.io"/>
--- a/pom.xml
+++ b/pom.xml
@@ -219,6 +219,28 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <!-- Ensure the manifest has all the gory details -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>3.0.2</version>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+                        </manifest>
+                    </archive>
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
@@ -261,6 +283,65 @@
     </reporting>
 
     <profiles>
+        <profile>
+            <!-- To be used during development. Run the command-->
+            <!-- mvn -Passembly package -->
+            <id>assembly</id>
+            <properties>
+                <skipTests>true</skipTests>
+                <checkstyle.ant.skip>true</checkstyle.ant.skip>
+                <checkstyle.skip>true</checkstyle.skip>
+                <pmd.skip>true</pmd.skip>
+                <findbugs.skip>true</findbugs.skip>
+                <xml.skip>true</xml.skip>
+                <forbiddenapis.skip>true</forbiddenapis.skip>
+                <cobertura.skip>true</cobertura.skip>
+                <maven.javadoc.skip>true</maven.javadoc.skip>
+                <linkcheck.skip>true</linkcheck.skip>
+                <!-- difference from "no-validations" -->
+                <maven.site.skip>true</maven.site.skip>
+            </properties>
+
+            <build>
+                <plugins>
+                    <!-- Creates the all inclusive uber jar -->
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-shade-plugin</artifactId>
+                        <version>3.0.0</version>
+                        <executions>
+                            <execution>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>shade</goal>
+                                </goals>
+                                <configuration>
+                                    <shadedArtifactAttached>true</shadedArtifactAttached>
+                                    <shadedClassifierName>all</shadedClassifierName>
+                                    <transformers>
+                                        <transformer
+                                                implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                            <mainClass>com.github.checkstyle.regression.Main</mainClass>
+                                        </transformer>
+                                    </transformers>
+                                    <filters>
+                                        <filter>
+                                            <artifact>*:*</artifact>
+                                            <excludes>
+                                                <exclude>META-INF/*.SF</exclude>
+                                                <exclude>META-INF/*.DSA</exclude>
+                                                <exclude>META-INF/*.RSA</exclude>
+                                            </excludes>
+                                        </filter>
+                                    </filters>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+
+        </profile>
         <profile>
             <id>cobertura-check</id>
             <activation>
--- a/src/main/java/com/github/checkstyle/regression/Main.java
+++ b/src/main/java/com/github/checkstyle/regression/Main.java
@@ -19,11 +19,51 @@
 
 package com.github.checkstyle.regression;
 
+import java.io.File;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.immutables.value.Value;
+
+import com.github.checkstyle.regression.configuration.ConfigGenerator;
+import com.github.checkstyle.regression.data.GitChange;
+import com.github.checkstyle.regression.data.ModuleExtractInfo;
+import com.github.checkstyle.regression.data.ModuleInfo;
+import com.github.checkstyle.regression.extract.ExtractInfoProcessor;
+import com.github.checkstyle.regression.git.DiffParser;
+import com.github.checkstyle.regression.module.ModuleCollector;
+import com.github.checkstyle.regression.module.ModuleUtils;
+import com.github.checkstyle.regression.report.ReportGenerator;
+
 /**
  * Utility class, contains main function and its auxiliary routines.
  * @author LuoLiangchen
  */
 public final class Main {
+    /** Option name of the local checkstyle repository path. */
+    private static final String OPT_LOCAL_GIT_REPO = "localGitRepo";
+
+    /** Option name of the PR branch name. */
+    private static final String OPT_PATCH_BRANCH = "patchBranch";
+
+    /** Option name of checkstyle-tester path. */
+    private static final String OPT_CHECKSTYLE_TESTER_PATH = "checkstyleTesterPath";
+
+    /** Option name of whether to stop after generating config. */
+    private static final String OPT_STOP_AFTER_GENERATE_CONFIG = "stopAfterGenerateConfig";
+
     /** Prevents instantiation. */
     private Main() {
     }
@@ -34,6 +74,168 @@ public final class Main {
      * @throws Exception execute failure
      */
     public static void main(String[] args) throws Exception {
-        // empty for now
+        final Options options = createOptions();
+        final CommandLineParser parser = new DefaultParser();
+        final HelpFormatter formatter = new HelpFormatter();
+        CommandLine cmd = null;
+
+        try {
+            cmd = parser.parse(options, args);
+        }
+        catch (ParseException ex) {
+            System.err.println(ex.getMessage());
+            formatter.printHelp("java -jar regression-tool.jar", options, true);
+            System.exit(1);
+        }
+
+        final Arguments arguments = ImmutableArguments.builder()
+                .repoPath(cmd.getOptionValue(OPT_LOCAL_GIT_REPO))
+                .branch(cmd.getOptionValue(OPT_PATCH_BRANCH))
+                .checkstyleTesterPath(
+                        Optional.ofNullable(cmd.getOptionValue(OPT_CHECKSTYLE_TESTER_PATH)))
+                .stopAfterGenerateConfig(cmd.hasOption(OPT_STOP_AFTER_GENERATE_CONFIG))
+                .build();
+
+        validateArguments(arguments);
+        runRegression(arguments);
+    }
+
+    /**
+     * Creates and initializes the {@link Options} instance.
+     * @return the initialized options
+     */
+    private static Options createOptions() {
+        final Options options = new Options();
+
+        final Option repo = Option.builder("r")
+                .longOpt(OPT_LOCAL_GIT_REPO)
+                .required()
+                .hasArg()
+                .desc("the path of the checkstyle repository")
+                .build();
+        repo.setRequired(true);
+        options.addOption(repo);
+
+        final Option branch = Option.builder("p")
+                .longOpt(OPT_PATCH_BRANCH)
+                .required()
+                .hasArg()
+                .desc("the name of the PR branch")
+                .build();
+        options.addOption(branch);
+
+        final Option tester = Option.builder("t")
+                .longOpt(OPT_CHECKSTYLE_TESTER_PATH)
+                .required(false)
+                .hasArg()
+                .desc("the path of the checkstyle-tester directory")
+                .build();
+        options.addOption(tester);
+
+        final Option stopAfterGenerateConfig = Option.builder()
+                .longOpt(OPT_STOP_AFTER_GENERATE_CONFIG)
+                .required(false)
+                .hasArg(false)
+                .desc("indicates that regression tool would stop after generating config")
+                .build();
+        options.addOption(stopAfterGenerateConfig);
+
+        return options;
+    }
+
+    /**
+     * Validates the parsed CLI arguments.
+     * @param args the parsed CLI arguments.
+     * @throws IllegalArgumentException the arguments are invalid
+     */
+    private static void validateArguments(Arguments args) {
+        if (!existAndIsDirectory(args.repoPath())) {
+            throw new IllegalArgumentException(
+                    "path of local git repo must exist and be a directory");
+        }
+        if (!args.stopAfterGenerateConfig()) {
+            if (args.checkstyleTesterPath().isPresent()) {
+                if (!existAndIsDirectory(args.checkstyleTesterPath().get())) {
+                    throw new IllegalArgumentException(
+                            "path of checkstyle tester must exist and be a directory");
+                }
+            }
+            else {
+                throw new IllegalArgumentException("missing checkstyleTesterPath, "
+                        + "which is required if you are not using --stopAfterGenerateConfig mode");
+            }
+        }
+    }
+
+    /**
+     * Runs the regression tool.
+     * @param args the parsed CLI arguments.
+     * @throws Exception execute failure
+     */
+    private static void runRegression(Arguments args) throws Exception {
+        final File config = generateConfig(args);
+        System.out.println("config generated at " + config.getAbsolutePath());
+        if (!args.stopAfterGenerateConfig()) {
+            final File report = ReportGenerator.generate(
+                    args.checkstyleTesterPath().get(), args.repoPath(), args.branch(), config);
+            System.out.println("report generated at " + report.getAbsolutePath());
+        }
+    }
+
+    /**
+     * Generates the config file.
+     * @param args the parsed CLI arguments.
+     * @return the generated config file
+     * @throws Exception generation failure
+     */
+    private static File generateConfig(Arguments args)
+            throws Exception {
+        final List<GitChange> changes = DiffParser.parse(args.repoPath(), args.branch());
+        final Map<String, ModuleExtractInfo> extractInfos =
+                ExtractInfoProcessor.getModuleExtractInfos(args.repoPath(), args.branch());
+        ModuleUtils.setNameToModuleExtractInfo(extractInfos);
+        final List<ModuleInfo> moduleInfos = ModuleCollector.generate(changes);
+        final DateFormat format = new SimpleDateFormat("yyyyMMddHHmmss", Locale.getDefault());
+        final String configFileName =
+                String.format("config-%s-%s.xml", args.branch(), format.format(new Date()));
+        return ConfigGenerator.generateConfig(configFileName, moduleInfos);
+    }
+
+    /**
+     * Checks whether the file in given path exists and is a directory.
+     * @param path the path to check
+     * @return true if file exists and is a directory
+     */
+    private static boolean existAndIsDirectory(String path) {
+        final File file = new File(path);
+        return !path.isEmpty() && file.exists() && file.isDirectory();
+    }
+
+    /** Represents the CLI arguments. */
+    @Value.Immutable
+    /* default */ interface Arguments {
+        /**
+         * The local checkstyle repository path.
+         * @return the local checkstyle repository path
+         */
+        String repoPath();
+
+        /**
+         * The PR branch name.
+         * @return the PR branch name
+         */
+        String branch();
+
+        /**
+         * Checkstyle-tester path.
+         * @return Checkstyle-tester path
+         */
+        Optional<String> checkstyleTesterPath();
+
+        /**
+         * Whether to stop after generating config.
+         * @return whether to stop after generating config
+         */
+        boolean stopAfterGenerateConfig();
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,12 @@ matrix:
         - CMD8="mvn test -Dtest=ExtractInfoGeneratorTest#generateExtractInfoFile"
         - CMD="$CMD1$CMD2$CMD3$CMD4$CMD5$CMD6$CMD7$CMD8"
 
+    # package no error (oraclejdk8)
+    - jdk: oraclejdk8
+      env:
+        - DESC="package no error"
+        - CMD="mvn clean package -Passembly"
+
 script: eval $CMD
 
 after_success:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -61,6 +61,10 @@ environment:
     - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
       DESC: "verify without checkstyle"
       CMD1: "mvn -e verify -Dcheckstyle.ant.skip=true -Dcheckstyle.skip=true"
+    # package no error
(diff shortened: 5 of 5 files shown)


rnveach:
> I failed to run the project by java -jar ... way.

I will create the configuration for the all jar.

> how should be present the exception message for user? 

I am fine throwing the exception all the way past `main`.

pom.xml
```diff
@@ -262,6 +262,55 @@
...
+                                <goals>
+                                    <goal>shade</goal>
+                                </goals>
+                                <configuration>
+                                    <shadedArtifactAttached>true</shadedArtifactAttached>
+                                    <shadedClassifierName>all</shadedClassifierName>
+                                    <transformers>
+                                        <transformer
+                                                implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                            <mainClass>com.puppycrawl.tools.checkstyle.Main</mainClass>
```
rnveach: Wrong package.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -19,11 +19,45 @@
...
+import com.github.checkstyle.regression.module.ModuleUtils;
+import com.github.checkstyle.regression.report.ReportGenerator;
+
 /**
  * Utility class, contains main function and its auxiliary routines.
  * @author LuoLiangchen
  */
 public final class Main {
+    /** Option name of repository path. */
+    private static final String OPT_REPO = "repo";
```
rnveach: Add short names and long names.

https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/diff.groovy#L44
Try to make similar options the same as the groovy scripts, that way there is less a learning curve.
This sound good?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -19,11 +19,45 @@
...
+
+    /** Option name of branch name. */
+    private static final String OPT_BRANCH = "branch";
+
+    /** Option name of checkstyle-tester path. */
+    private static final String OPT_TESTER = "tester";
```
rnveach: contribution is the name of the repo. checkstyle-tester is the name of the folder.
Can we choose a better, more precise name?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +68,91 @@ private Main() {
...
     public static void main(String[] args) throws Exception {
-        // empty for now
+        final Options options = createOptions();
+        final CommandLineParser parser = new DefaultParser();
+        final HelpFormatter formatter = new HelpFormatter();
+        final CommandLine cmd;
+
+        try {
+            cmd = parser.parse(options, args);
+            final String repoPath = cmd.getOptionValue(OPT_REPO);
```
rnveach: All paths and other fields need to be validated if they are entered.
Paths must exist and be a directory, etc..
etc..

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +68,91 @@ private Main() {
...
+
+            final File config = generateConfig(repoPath, branch);
+            System.out.println("config generated at " + config.getAbsolutePath());
+            if (!stopAfterGenerateConfig) {
+                final String testerPath = cmd.getOptionValue(OPT_TESTER);
+                final File report = ReportGenerator.generate(testerPath, repoPath, branch, config);
+                System.out.println("report generated at " + report.getAbsolutePath());
+            }
+        }
+        catch (ParseException ex) {
```
rnveach: Let's just throw exception past `main`.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +68,91 @@ private Main() {
...
+     */
+    private static File generateConfig(String repoPath, String branch)
+            throws Exception {
+        final List<GitChange> changes = DiffParser.parse(repoPath, branch);
+        final Map<String, ModuleExtractInfo> extractInfos =
+                ExtractInfoProcessor.getModuleExtractInfos(repoPath, branch);
+        ModuleUtils.setNameToModuleExtractInfo(extractInfos);
+        final List<ModuleInfo> moduleInfos = ModuleCollector.generate(changes);
+        final String configFileName =
+                String.format("config-%s-%d.xml", branch, System.currentTimeMillis());
```
rnveach: If we are going to format it, can we do date and time instead of milliseconds?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +68,91 @@ private Main() {
...
+        final HelpFormatter formatter = new HelpFormatter();
+        final CommandLine cmd;
+
+        try {
+            cmd = parser.parse(options, args);
+            final String repoPath = cmd.getOptionValue(OPT_REPO);
+            final String branch = cmd.getOptionValue(OPT_BRANCH);
+            final boolean stopAfterGenerateConfig = cmd.hasOption(OPT_STOP_AFTER_GENERATE_CONFIG);
+
+            final File config = generateConfig(repoPath, branch);
```
rnveach: this and lines below it should all be in `runRegression`.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: So `checkstyleTester`?

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: What is `runRegression`? Do you mean create a new method with name `runRegression`?

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: I think it is OK to throw exceptions from components. 

But this is exactly a argument format exception, it is only caused by user input 1. wrong format 2.missing arguments 3. wrong argument number etc. And in this case we need to print help message(by `formatter.printHelp`) and tell user how to use this tool. So I think this catch is necessary.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: I made this changes:
`repo` => `localGitRepo`, and short opt is `r`
`branch` => `patchBranch` and short opt is `p`

Do we really need a parameter of `baseBranch`, is there any possible that it is not `master`?

The option name of checkstyle-tester path is being discussed in the post below.

Short opts are defined in `createOptions` directly. Need I create static final field instead of writing them in plain text in `createOptions`?

src/main/java/com/github/checkstyle/regression/Main.java
rnveach: > 1. wrong format 2.missing arguments 3. wrong argument number etc.

Than why is it covering everything if it is only for user input? What lines are throwing this?

src/main/java/com/github/checkstyle/regression/Main.java
rnveach: This is done.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +72,110 @@ private Main() {
...
+            cmd = parser.parse(options, args);
+            final String repoPath = cmd.getOptionValue(OPT_LOCAL_GIT_REPO);
+            if (!existAndIsDirectory(repoPath)) {
+                throw new IllegalArgumentException(
+                        "path of local git repo must exist and be a directory");
+            }
+            final String branch = cmd.getOptionValue(OPT_PATCH_BRANCH);
+            final boolean stopAfterGenerateConfig = cmd.hasOption(OPT_STOP_AFTER_GENERATE_CONFIG);
+
+            final File config = generateConfig(repoPath, branch);
```
rnveach: > What is runRegression? Do you mean create a new method with name runRegression?

yes, and move these liens into it.
I want separation between validating input and running of regression.
Similar to https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L381 and https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L287 .

config/import-control.xml
```diff
@@ -4,6 +4,9 @@
         "http://www.puppycrawl.com/dtds/import_control_1_1.dtd">
 
 <import-control pkg="com.github.checkstyle.regression">
+    <allow pkg="org.apache.commons.cli" local-only="true"/>
+    <allow pkg="com.github.checkstyle.regression" local-only="true"/>
+    <allow pkg="java.text"/>
```
rnveach: Let's make this local only too for now.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: The first line. I will update this.

config/import-control.xml

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: Done. Add a `validateArguments` and `runRegression`

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: Updated. Only the parsing line in try block now.

Luolc:
> We need an issue on code coverage in POM.

#58 

> checkstyle-tester is the name of the folder.
Can we choose a better, more precise name?

Now it is named as `checkstyleTester`. Is that ok?

I forget to reply done to some of the reviews before, sorry.

src/main/java/com/github/checkstyle/regression/Main.java
rnveach: > Do we really need a parameter of baseBranch? 

No. I was only recommending changes for existing parameters.

> Need I create static final field instead of writing them in plain text in createOptions?

I am fine without this for now.

rnveach:
> Now it is named as checkstyleTester. Is that ok?

Shouldn't it end with path to better signify we want the path?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+     * Validates the parsed CLI arguments.
+     * @param args the parsed CLI arguments.
+     * @throws IllegalArgumentException the arguments are invalid
+     */
+    private static void validateArguments(Arguments args) {
+        if (!existAndIsDirectory(args.repoPath())) {
+            throw new IllegalArgumentException(
+                    "path of local git repo must exist and be a directory");
+        }
+        if (!args.stopAfterGenerateConfig() && !existAndIsDirectory(args.testerPath().get())) {
```
rnveach: If I don't give any value for `testerPath`, will `args.testerPath()` return `null`?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+        catch (ParseException ex) {
+            System.err.println(ex.getMessage());
+            formatter.printHelp(null, options);
+            System.exit(1);
+        }
+
+        final Arguments arguments = ImmutableArguments.builder()
+                .repoPath(cmd.getOptionValue(OPT_LOCAL_GIT_REPO))
+                .branch(cmd.getOptionValue(OPT_PATCH_BRANCH))
+                .testerPath(cmd.getOptionValue(OPT_CHECKSTYLE_TESTER))
```
rnveach: please use full name for this one, `testerPath`. It just seems slightly off.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+     * @throws IllegalArgumentException the arguments are invalid
+     */
+    private static void validateArguments(Arguments args) {
+        if (!existAndIsDirectory(args.repoPath())) {
+            throw new IllegalArgumentException(
+                    "path of local git repo must exist and be a directory");
+        }
+        if (!args.stopAfterGenerateConfig() && !existAndIsDirectory(args.testerPath().get())) {
+            throw new IllegalArgumentException(
+                    "path of checkstyle tester must exist and be a directory");
```
rnveach: if I don't give tester path, this statement will look odd as it doesn't also say it is a required field.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+
+    /**
+     * Validates the parsed CLI arguments.
+     * @param args the parsed CLI arguments.
+     * @throws IllegalArgumentException the arguments are invalid
+     */
+    private static void validateArguments(Arguments args) {
+        if (!existAndIsDirectory(args.repoPath())) {
+            throw new IllegalArgumentException(
+                    "path of local git repo must exist and be a directory");
```
rnveach: if I don't give repo path, this statement will look odd as it doesn't also say it is a required field.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+    }
+
+    /**
+     * Generates the config file.
+     * @param repoPath the path of the checkstyle repository
+     * @param branch   the name of the PR branch
+     * @return the generated config file
+     * @throws Exception generation failure
+     */
+    private static File generateConfig(String repoPath, String branch)
```
rnveach: move closer to `runRegression`. it is like a child method.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+        final List<ModuleInfo> moduleInfos = ModuleCollector.generate(changes);
+        final DateFormat format = new SimpleDateFormat("yyyyMMddHHmmss", Locale.getDefault());
+        final String configFileName =
+                String.format("config-%s-%s.xml", branch, format.format(new Date()));
+        return ConfigGenerator.generateConfig(configFileName, moduleInfos);
+    }
+
+    /** Represents the CLI arguments. */
+    @Value.Immutable
+    /* default */ interface Arguments {
```
rnveach: why the `default` comment?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,160 @@ private Main() {
...
+        }
+    }
+
+    /**
+     * Checks whether the file in given path exists and is a directory.
+     * @param path the path to check
+     * @return true if file exists and is a directory
+     */
+    private static boolean existAndIsDirectory(String path) {
+        final File file = new File(path);
```
rnveach: if `path` can be `null` or empty string, won't this incorrectly return that it exists?
verify first that `null` or empty string can get into here.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: `Optional.get()` would throw an exception if we don't have a value. http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/Optional.java#Optional.get%28%29

I would update this code.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: It can't be null, but could be empty if we write smth like `path=""` in CLI.

Check is added.

Luolc:
@rnveach 
> > Now it is named as checkstyleTester. Is that ok?

> Shouldn't it end with path to better signify we want the path?

Update to `checkstyleTesterPath`

> if I don't give repo path, this statement will look odd as it doesn't also say it is a required field.
if I don't give tester path, this statement will look odd as it doesn't also say it is a required field.

As we init the option properties in `createOptions`, `localGitRepo` and `patchBranch` is required, `checkstyleTesterPath` and `stopAfterGenerateConfig` is not.
If `localGitRepo` is missing, then program would stop at
```java
        try {
            cmd = parser.parse(options, args);
        }
        catch (ParseException ex) {
            System.err.println(ex.getMessage());
            formatter.printHelp(null, options);
            System.exit(1);
        }
```
as there would be a `ParseException`, with message like `Missing required option XX`.
Meanwhile, `immutable` class would validate fields not null when it is being created. So `repoPath` is impossible to be null in `validateArguments` method.

But `checkstyleTesterPath` is not required, so we use `Optional` field in `Arguments`. And a `NoSuchElementException` would be thrown if we try to do `checkstyleTesterPath().get()` when it is null.
I have updated `validateArguments` to throw a clearer message for this case.

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: Would have a PMD failure without this. `Rule:CommentDefaultAccessModifier`
https://pmd.github.io/pmd-5.4.0/pmd-java/rules/java/comments.html#CommentDefaultAccessModifier

We need `package-private` here. `private` and `protected` can't be inherited by `immutables` and `public` is not necessary. 
We have another choice that updating pmd.xml to ignore this, instead of this comment.

rnveach: ok, ignore this then. we will leave it for now.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,169 @@ private Main() {
...
+                throw new IllegalArgumentException("missing checkstyleTesterPath, "
+                        + "which is required if you are using --stopAfterGenerateConfig mode");
+            }
+        }
+    }
+
+    /**
+     * Runs the regression tool.
+     * @param args the parsed CLI arguments.
+     * @throws Exception excute failure
```
rnveach: `excute` => `execute`

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,169 @@ private Main() {
...
+    }
+
+    /**
+     * Generates the config file.
+     * @param repoPath the path of the checkstyle repository
+     * @param branch   the name of the PR branch
+     * @return the generated config file
+     * @throws Exception generation failure
+     */
+    private static File generateConfig(String repoPath, String branch)
```
rnveach: Even though we don't need the entire arguments class, I sorta like sending it better than its pieces. Less documentation and easier to grab new arguments that we add.
What do you think about this?

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: Done. My typo check was accidentally disabled before. :-(

src/main/java/com/github/checkstyle/regression/Main.java
Luolc: My mistake. I forgot to update this one.
This is done.

Luolc:
@rnveach 
> Please add to issue what was done and what was left when this is merged. Don't forget we need full documentation in the readme.

I am facing an error `Exception in thread "main" java.lang.SecurityException: Invalid signature file digest for Manifest main attributes`, when trying `java -jar ..` to run the jar file. I have searched about this for a while but failed to solve it. By now I haven't really use CLI to see how it works now.

Docs in other issue?

rnveach:
> Exception in thread "main" java.lang.SecurityException: Invalid signature file digest for Manifest main attributes

I will look into. Please add JAR generation to travis and appveyor.

rnveach:
> Manifest 

Try adding this: https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L927-L947

Luolc:
@rnveach 
> Try adding this: https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L927-L947
Still having error. pom update is pushed.

My command:
```bash
$ mvn clean package -Passembly
$ java -jar target/regression-tool-1.0-SNAPSHOT-all.jar -r path/to/checkstyle -b branch --stopAfterGenerateConfig
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.SecurityException: Invalid signature file digest for Manifest main attributes
	at sun.security.util.SignatureFileVerifier.processImpl(SignatureFileVerifier.java:314)
	at sun.security.util.SignatureFileVerifier.process(SignatureFileVerifier.java:268)
	at java.util.jar.JarVerifier.processEntry(JarVerifier.java:316)
	at java.util.jar.JarVerifier.update(JarVerifier.java:228)
	at java.util.jar.JarFile.initializeVerifier(JarFile.java:383)
	at java.util.jar.JarFile.getInputStream(JarFile.java:450)
	at sun.misc.URLClassPath$JarLoader$2.getInputStream(URLClassPath.java:940)
	at sun.misc.Resource.cachedInputStream(Resource.java:77)
	at sun.misc.Resource.getByteBuffer(Resource.java:160)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:454)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:495)
```

> Please add JAR generation to travis and appveyor.

Do you mean adding a `mvn clean package -Passembly` in travis and appveyor?

rnveach:
> Do you mean adding a mvn clean package -Passembly in travis and appveyor?

Yes.

rnveach:
@Luolc see http://zhentao-li.blogspot.com/2012/06/maven-shade-plugin-invalid-signature.html . It is the first result that comes up in google and it fixed the issue for me.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,168 @@ private Main() {
...
+        final CommandLineParser parser = new DefaultParser();
+        final HelpFormatter formatter = new HelpFormatter();
+        CommandLine cmd = null;
+
+        try {
+            cmd = parser.parse(options, args);
+        }
+        catch (ParseException ex) {
+            System.err.println(ex.getMessage());
+            formatter.printHelp(null, options);
```
rnveach: This produces the following exception when called with no arguments:
````
java.lang.IllegalArgumentException: cmdLineSyntax not provided
````

Comment (Luolc) [inline src/main/java/com/github/checkstyle/regression/Main.java]: Fixed. `cmdLineSyntax` is not allowed to be null or empty. Now I pass a `regression-tool` to it. (the argument is shown after `usage:`)
Acts like:
```bash
$ java -jar target/regression-tool-1.0-SNAPSHOT-all.jar
Missing required options: r, p
usage: regression-tool
 -p,--patchBranch <arg>            the name of the PR branch
 -r,--localGitRepo <arg>           the path of the checkstyle repository
    --stopAfterGenerateConfig      indicates that regression tool would
                                   stop after generating config
 -t,--checkstyleTesterPath <arg>   the path of the checkstyle-tester
                                   directory
```

Luolc:
> Please add JAR generation to travis and appveyor.

Done.

> see http://zhentao-li.blogspot.com/2012/06/maven-shade-plugin-invalid-signature.html . It is the first result that comes up in google and it fixed the issue for me.

It works, thanks. I found similar answer at StackOverflow before but it seems missed line `<artifact>*:*</artifact>`.

rnveach:
> usage: regression-tool

It should be `java -jar regression-tool.jar` (not sure about version number) and shouldn't it include some text describing the optional arguments like other programs do?
Example from java:
````
Usage: java [-options] class [args...]
          (to execute a class)
  or  java [-options] -jar jarfile [args...]
          (to execute a jar file)
````

**Comment by Luolc** [comment]
@rnveach 
I wrote `regression-tool` for I read its source code and found the parameter of its internal method is named as `appName`, while I don't actually know why.

 **Edit:** I find the proper way in javadoc of class `HelpFormatter`.
```java
formatter.printHelp("myapp", header, options, footer, true);
```
Sample output:
```bash
usage: myapp -f <FILE> [-h] [-v]
```
Now I understand the meaning of the first parameter. It supposed we are running command like `mvn ...`, `git ...` etc. And the last parameter `true` is turning on `autoUsage` mode.

Now our code:
```java
formatter.printHelp("java -jar regression-tool.jar", options, true);
```

Output:
```bash
$ java -jar target/regression-tool-1.0-SNAPSHOT-all.jar
Missing required options: r, p
usage: java -jar regression-tool.jar -p <arg> -r <arg>
       [--stopAfterGenerateConfig] [-t <arg>]
 -p,--patchBranch <arg>            the name of the PR branch
 -r,--localGitRepo <arg>           the path of the checkstyle repository
    --stopAfterGenerateConfig      indicates that regression tool would
                                   stop after generating config
 -t,--checkstyleTesterPath <arg>   the path of the checkstyle-tester
                                   directory
```

One problem is the version number. Should we specify it manually every time we update, or trying to detect the version number from maven settings?

rnveach:
@Luolc lets leave version number to another time. It will have to detect snapshot too. There should be some way for maven to inject the needed data for us to acquire at runtime.

Luolc:
@rnveach OK. 
Is there anything else need to improve in this PR?

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,168 @@ private Main() {
...
+        final CommandLineParser parser = new DefaultParser();
+        final HelpFormatter formatter = new HelpFormatter();
+        CommandLine cmd = null;
+
+        try {
+            cmd = parser.parse(options, args);
+        }
+        catch (ParseException ex) {
+            System.err.println(ex.getMessage());
+            formatter.printHelp("regression-tool", options);
```
rnveach: waiting for this update...

Luolc: Sorry I found I didn't successfully push the changes before. Updated.

src/main/java/com/github/checkstyle/regression/Main.java
```diff
@@ -34,6 +74,168 @@ private Main() {
...
+        if (!args.stopAfterGenerateConfig()) {
+            if (args.checkstyleTesterPath().isPresent()) {
+                if (!existAndIsDirectory(args.checkstyleTesterPath().get())) {
+                    throw new IllegalArgumentException(
+                            "path of checkstyle tester must exist and be a directory");
+                }
+            }
+            else {
+                throw new IllegalArgumentException("missing checkstyleTesterPath, "
+                        + "which is required if you are using --stopAfterGenerateConfig mode");
```
rnveach: shouldn't this be `if you are not using --stopAfterGenerateConfig mode`?
The if statement above is `if (!args.stopAfterGenerateConfig())`.

Luolc: Done.

Luolc:
@rnveach 
> What about this comment at https://github.com/checkstyle/regression-tool/pull/57/files/6a2f57678ca9ae6ab78bb5cd7b5669200be3f45e#discussion_r128928723 ?

I got a `We went looking everywhere, but couldn’t find those commits.` page.

rnveach:
@Luolc I edited the comment after I noticed it, please re-check the new link.

Luolc:
@rnveach
I mentioned this at https://github.com/checkstyle/regression-tool/pull/57#issuecomment-317334010

The explanation in detail is at that post.
If repo path is not given, the program would stop earlier and at that line repo is not possible to be null.

rnveach:
ok, I missed that. Once CI is done I will merge.

Luolc:
#59 This issue for creating a document of this tool.

PR #61, opened 2017-07-28
Issue #15: NoException Checkstyle Project Regression

Luolc:
#15 

Current one is only part of the test.
I cloned checkstyle project and run with `--stopAfterGenerateConfig`.
Assumed chekcstyle has commis as following:
```
commit1 (HEAD)
commit2
commit3
commit4
...
```
I checkout to a new branch `test-branch` with commit1 as HEAD, and revert master to commit2, then do the test.
I could also create other pairs, and I think there are two ways:
- commit1(branch), commit2(master); commit2(branch), commit3(master); commit3(branch), commit4(master)...
- commit1(branch), commit2(master); commit1(branch), commit3(master); commit1(branch), commit4(master)...

Which one do you think it better? I prefer the first now.

About checkstyle-tester, we need also clone it and maybe need to prepare a `project-to-test` for this test. Do you think the default one in contribution repo is OK?

Also, I am considering to manage the command in a independent file like `.ci/travis` in checkstyle project. This test need a lot of command later.

PR #70, opened 2017-08-09
Issue #15: NoException Checkstyle Project Regression

rnveach:
https://github.com/checkstyle/regression-tool/issues/15

Continuation of https://github.com/checkstyle/regression-tool/pull/61 .

### DIFF
--- a/.ci/travis/checkstyle_regression_no_exception.sh
+++ b/.ci/travis/checkstyle_regression_no_exception.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+BRANCH_COMMIT="$1"
+MASTER_COMMIT="$2"
+CHECKSTYLE_VERSION="$3"
+
+if [ ! -d checkstyle ]; then
+    git clone https://github.com/checkstyle/checkstyle
+fi
+
+if [ ! -d contribution ]; then
+    git clone --depth=1 https://github.com/checkstyle/contribution
+fi
+if [ ! -z "$CHECKSTYLE_VERSION" ]; then
+    sed -i'' "s/<checkstyle\.version>.*<\/checkstyle\.version>/<checkstyle\.version>${CHECKSTYLE_VERSION}<\/checkstyle\.version>/g" \
+    contribution/checkstyle-tester/pom.xml
+fi
+
+cd checkstyle
+CHECKSTYLE_PATH="$(pwd)"
+git branch test-branch
+git checkout test-branch
+git reset --hard "${BRANCH_COMMIT}"
+git checkout master
+git reset --hard "${MASTER_COMMIT}"
+cd ..
+mvn clean package -Passembly
+java -jar target/regression-tool-1.0-SNAPSHOT-all.jar -r "$CHECKSTYLE_PATH" -p test-branch -t contribution/checkstyle-tester
+RESULT="$?"
+cat config-test-branch-*.xml
+rm config-test-branch-*.xml
+
+if [ "$?" != "0" ] || [ "$RESULT" != "0" ]; then
+    exit "1";
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: java
 
 sudo: false
+dist: precise
 
 before_install:
   - export M2_HOME=/usr/local/maven
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: java
 
 sudo: false
 
+before_install:
+  - export M2_HOME=/usr/local/maven
+
 install:
   -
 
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -66,6 +69,14 @@ matrix:
         - DESC="package no error"
         - CMD="mvn clean package -Passembly"
 
+    # checkstyle regression no exception on latest commit (oraclejdk8)
+    - jdk: oraclejdk8
+      env:
+        - DESC="checkstyle regression no exception on latest commit"
+        - CMD1="./.ci/travis/checkstyle_regression_no_exception.sh HEAD HEAD^ &&"
+        - CMD2="./.ci/travis/checkstyle_regression_no_exception.sh 44efd23 4b66933 8.1-SNAPSHOT"
```
rnveach: needs comment about what is being tested.

.ci/travis/checkstyle_regression_no_exception.sh
```diff
@@ -0,0 +1,35 @@
...
+cd checkstyle
+CHECKSTYLE_PATH="$(pwd)"
+git branch test-branch
+git checkout test-branch
+git reset --hard "${BRANCH_COMMIT}"
+git checkout master
+git reset --hard "${MASTER_COMMIT}"
+cd ..
+mvn clean package -Passembly
+java -jar target/regression-tool-1.0-SNAPSHOT-all.jar -r "$CHECKSTYLE_PATH" -p test-branch -t contribution/checkstyle-tester
```
rnveach: needs to check return code for failure.

rnveach:
I'm stopping the build. I still think it is the distribution setting on travis as it is taking too long.

rnveach:
> Error: Could not find or load main class org.codehaus.plexus.classworlds.launcher.Launcher

M2_HOME is not correct for all travis.
Original travis maven was in:
> Maven home: /usr/local/maven-3.3.9

Current travis maven is in:
> Maven home: /usr/local/maven

I will try hardcoding it but we probably should set it dynamically.

rnveach:
@Luolc travis passed now, it was because of the missing distribution. We saw something similar in Checkstyle.
I could have sworn I asked you to try adding it, but I couldn't find anything.

.ci/travis/checkstyle_regression_no_exception.sh
rnveach: This is already done.

.travis.yml
rnveach: Done.

rnveach:
I split the travis into 2 so we can work them in parallel.
I changed contribution cloning to not constantly delete the directory, as it should work with successive runs.

rnveach:
Let me know if you see anything we should change before we merge.

Luolc:
@rnveach 
> I could have sworn I asked you to try adding it, but I couldn't find anything.

Sorry I misunderstood your commend, I was thinking that you mean to update my local checkstyle repo.

Actually I was testing on my local by just running `./.ci/travis/checkstyle_regression_no_exception.sh 44efd23 4b66933 8.1-SNAPSHOT`. The Travis settings shouldn't affect the result. I still got the error on my local.

.ci/travis/checkstyle_regression_no_exception.sh
```diff
@@ -0,0 +1,34 @@
...
+BRANCH_COMMIT="$1"
+MASTER_COMMIT="$2"
+CHECKSTYLE_VERSION="$3"
+
+if [ ! -d checkstyle ]; then
+    git clone https://github.com/checkstyle/checkstyle
+fi
+
+if [ ! -d contribution ]; then
+    git clone --depth=1 https://github.com/checkstyle/contribution
```
Luolc: At the beginning I wrote like this as well, but found some problems and changed it to delete the repo every time. But I can't remember the reason.
Now there is only one test with specific commit id. Maybe there would be problems when we test twice(HEAD and specific)

.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: java
 
 sudo: false
+dist: precise
+
+before_install:
+  - export M2_HOME=/usr/local/maven
```
Luolc: We may install maven manually instead of using the default one. But disadvantage is that we can only use maven with a specific version. Do you think it necessary?

rnveach: Manually installing maven, no. But maven command exists, we should be able to do somrthing like `export M2_HOME=$(which mvn)`.

.ci/travis/checkstyle_regression_no_exception.sh
rnveach: > Maybe the cause is testing HEAD and specific commit in one time.

That can be an issue for `checkstyle/checkstyle`. It shouldn't be an issue for contribution which we don't switch branches on.

rnveach:
Since travis passed I will merge.

> I was testing on my local by just running ./.ci/travis/checkstyle_regression_no_exception.sh 44efd23 4b66933 8.1-SNAPSHOT.  I still got the error on my local.

Please confirm if this is still a problem.
If it is, make an issue for it as someone else could have this problem too.

Issue #67, opened 2017-08-04
Users should be able to run the tool in any directory


Luolc:
Taken from https://github.com/checkstyle/regression-tool/pull/60#discussion_r130597286

Now the tool could be only run in the same directory as source code.
We need to fix this and update documentation.

PR #72, opened 2017-08-10
Issue #67: Users should be able to run the tool in any directory

Luolc:
#67 

Maybe we need to add a test of running outside source directory in Travis after #15 is done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,6 @@ In most case, it should be your forked repository.
 - Regression-tool uses environment variable `M2_HOME` or system property `maven.home` to find the maven binary files.
 You need to specify either `M2_HOME` or `maven.home` in advance. You could get the path of maven home by running `mvn -v`.
 
-> Temporarily, the tool must be run in the same directory as source code. This would be fix soon.
-
 ### Clone
 
 Just clone this repository to your local.
@@ -93,6 +91,6 @@ $ java -jar target/regression-tool-XX-all.jar -r /path/to/checkstyle/ -p issue12
 
 ### Output
 
-The config file would be generated in `regression-tool` directory.
+The config file would be generated in current working directory.
 
 The report would be generated somewhere in TBD.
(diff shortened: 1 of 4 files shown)


rnveach:
@Luolc #15 is done.
I don't feel running outside the directory needs to be added to CI. CI script must use specific directory as it is to compile package.
Please confirm locally that it now works outside source directory.

README.md
```diff
@@ -93,6 +91,6 @@ $ java -jar target/regression-tool-XX-all.jar -r /path/to/checkstyle/ -p issue12
 
 ### Output
 
-The config file would be generated in `regression-tool` directory.
+The config file would be generated in current directory.
```
rnveach: `current` => `current working`

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
```diff
@@ -139,14 +139,13 @@ private void invokeMavenCommand() throws InjectException {
      * @throws IOException failure of copying
      */
     private void copyInjectFilesToCheckstyleRepo() throws IOException {
-        final File srcDir =
-                new File("src/main/resources/com/github/checkstyle/regression/extract/");
+        final String path = "/com/github/checkstyle/regression/extract/";
         final File destDir = new File(repoPath, "src/test/java/com/puppycrawl/tools/checkstyle/");
         final String[] injections = {"ExtractInfoGeneratorTest.java", "JsonUtil.java"};
```
rnveach: We probably should have a test to verify all files in extract directory are listed in this string.
Please make a new issue.

README.md
Luolc: Done.

src/main/java/com/github/checkstyle/regression/extract/CheckstyleInjector.java
Luolc: #74

Luolc:
@rnveach 
> Please confirm locally that it now works outside source directory

Tested on my local. Successful.

Issue #77, opened 2017-08-16
git: add line changes detection


Luolc:
We need to update `GitChange`. Add a field to store the line numbers of the changed code.
And update `DiffParser` to let it be able to detect line changes.

rnveach:
> the number of changed lines.

Why do we need the number of changed lines instead the line numbers that changed?
We will need exact line numbers for things like Issue #69 .

Luolc:
> Why do we need the number of changed lines instead of the line numbers that changed?

My expression was wrong. I updated the first post.

rnveach:
Fix was merged

Issue #69, opened 2017-08-09
Recognize difference between comments and code changes


rnveach:
Example commit: https://github.com/checkstyle/checkstyle/commit/2a5654734a64c5459819c5ced97bdacd0285ccc7

Lots of no-inspection comments were added, only a few actual code changes but regression was still run on all checks in commit.

Generated configuration:
````
  <module name="TreeWalker">
    <module name="IndentationCheck"/>
    <module name="CyclomaticComplexityCheck"/>
  </module>
  <module name="UniquePropertiesCheck"/>
````

Only `CyclomaticComplexityCheck` had actual code changes and needs regression.

Issue #49, opened 2017-07-06
DiffParser: ignore chmod change


Luolc:
Taken from https://github.com/checkstyle/regression-tool/pull/22#discussion_r125713238

If the only change is a chmod(no changes to the contents of the file), we should ignore the change altogether.

After this if fixed, the UT of only chmod change in #22 should update.

We will need a new test for this: chmod plus file change.

Luolc:
Some experiment:
```bash
$ git init
$ echo hello > HelloWorld
$ git add .
$ git commit -m "initial commit"
[master (root-commit) 83e0652] initial commit
 1 file changed, 1 insertion(+)
 create mode 100644 HelloWorld
$
$ git checkout -b foo
Switched to a new branch 'foo'
$ chmod 755 HelloWorld
$ git status
On branch foo
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

  modified:   HelloWorld

no changes added to commit (use "git add" and/or "git commit -a")
$ git add .
$ git commit -m "change filemode"
[foo f3c937d] change filemode
 1 file changed, 0 insertions(+), 0 deletions(-)
 mode change 100644 => 100755 HelloWorld
$
$ git diff master
diff --git a/HelloWorld b/HelloWorld
old mode 100644
new mode 100755
$ git config core.filemode false
$ git diff master
diff --git a/HelloWorld b/HelloWorld
old mode 100644
new mode 100755
$
$ chmod 644 HelloWorld
$ git status
On branch foo
nothing to commit, working tree clean
```

So we can't ignore filemode change after a mode change is already committed. The only way I could find now is checking this by the size of "edit lines". If the size is zero, then this is only a filemode change.

IMO, we'd better work on this after we introducing line changes in the future.

rnveach:
Fix was merged

Issue #6, opened 2017-06-05
Utility Regression


rnveach:
We have many utilities that our modules rely on.
Example: https://github.com/checkstyle/checkstyle/tree/master/src/main/java/com/puppycrawl/tools/checkstyle/utils

If we make a change to them, we need to run regression on the modules that rely on them.

First we need to identify which methods in the utilities were changed. 
Then we need to identify which modules are affected by looking at import statements and then looking for those methods to be called in the module.
We have to beware of utilities that call other utilities as we need to be able to trace all of them back to their modules that use them.

This is blocked by https://github.com/checkstyle/regression-tool/issues/5 .

PR #82, opened 2017-08-16
Issue #77: git: add line changes detection

Luolc:
#77

In method `parse`, we have to use for-loop instead of `map`, for `convertDiffEntryToGitChange` now throws `IOException`.

Since `IntStream.range` is banned for the cobertura bug, I use apache.common's `IntRange` instead.

I add a comment to ignore `ClassDataAbstractionCoupling`.

### DIFF
--- a/src/main/java/com/github/checkstyle/regression/data/GitChange.java
+++ b/src/main/java/com/github/checkstyle/regression/data/GitChange.java
@@ -19,6 +19,8 @@
 
 package com.github.checkstyle.regression.data;
 
+import java.util.List;
+
 import org.immutables.value.Value;
 
 /**
@@ -32,4 +34,18 @@ public interface GitChange {
      * @return the path of the changed file
      */
     String path();
+
+    /**
+     * The line numbers of the added changes.
+     * The first line of a file is marked as line zero.
+     * @return the line numbers of the added changes
+     */
+    List<Integer> addedLines();
+
+    /**
+     * The line numbers of the deleted changes.
+     * The first line of a file is marked as line zero.
+     * @return the line numbers of the deleted changes
+     */
+    List<Integer> deletedLines();
 }
--- a/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
+++ b/src/main/java/com/github/checkstyle/regression/git/DiffParser.java
@@ -21,12 +21,16 @@ package com.github.checkstyle.regression.git;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.math.IntRange;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.diff.DiffEntry;
+import org.eclipse.jgit.diff.DiffFormatter;
 import org.eclipse.jgit.lib.Constants;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.RevCommit;
@@ -36,6 +40,7 @@ import org.eclipse.jgit.revwalk.filter.RevFilter;
 import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
 import org.eclipse.jgit.treewalk.AbstractTreeIterator;
 import org.eclipse.jgit.treewalk.CanonicalTreeParser;
+import org.eclipse.jgit.util.io.DisabledOutputStream;
 
 import com.github.checkstyle.regression.data.GitChange;
 import com.github.checkstyle.regression.data.ImmutableGitChange;
@@ -44,6 +49,7 @@ import com.github.checkstyle.regression.data.ImmutableGitChange;
  * Parses git diff between PR branch and master for the further use.
  * @author LuoLiangchen
  */
+// -@cs[ClassDataAbstractionCoupling] We have to import many classes from JGit
 public final class DiffParser {
     /** Prevents instantiation. */
     private DiffParser() {
@@ -59,7 +65,7 @@ public final class DiffParser {
      */
     public static List<GitChange> parse(String repositoryPath, String branchName)
             throws IOException, GitAPIException {
-        final List<GitChange> returnValue;
+        final List<GitChange> returnValue = new LinkedList<>();
         final File gitDir = new File(repositoryPath, ".git");
         final Repository repository = new FileRepositoryBuilder().setGitDir(gitDir)
                 .readEnvironment().findGitDir().build();
@@ -67,16 +73,20 @@ public final class DiffParser {
         try {
             final TreeParserPair pair = getTreeParserPair(repository, branchName);
             final Git git = new Git(repository);
+            final DiffFormatter formatter = new DiffFormatter(DisabledOutputStream.INSTANCE);
+            formatter.setRepository(repository);
 
             try {
-                returnValue = git.diff()
+                final List<DiffEntry> diffs = git.diff()
                         .setOldTree(pair.commonAncestorTreeParser)
                         .setNewTree(pair.prTreeParser)
                         .call()
                         .stream()
                         .filter(entry -> entry.getChangeType() != DiffEntry.ChangeType.DELETE)
-                        .map(DiffParser::convertDiffEntryToGitChange)
                         .collect(Collectors.toList());
+                for (DiffEntry diff : diffs) {
+                    returnValue.add(convertDiffEntryToGitChange(diff, formatter));
+                }
             }
             finally {
                 git.close();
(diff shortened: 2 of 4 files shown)


src/main/java/com/github/checkstyle/regression/git/DiffParser.java
```diff
@@ -59,24 +65,28 @@ private DiffParser() {
...
 
             try {
-                returnValue = git.diff()
+                final List<DiffEntry> diffs = git.diff()
                         .setOldTree(pair.commonAncestorTreeParser)
                         .setNewTree(pair.prTreeParser)
                         .call()
                         .stream()
                         .filter(entry -> entry.getChangeType() != DiffEntry.ChangeType.DELETE)
-                        .map(DiffParser::convertDiffEntryToGitChange)
```
rnveach: this can't be made into a lambda?
````
diff -> convertDiffEntryToGitChange(diff, formatter)
````

File: src/main/java/com/github/checkstyle/regression/git/DiffParser.java [hunk 2]
```diff
@@ -158,11 +168,20 @@ private static AbstractTreeIterator prepareTreeParser(RevWalk walk, RevCommit co
...
+     * @param formatter the diff formatter to provide the line changes information
      * @return the {@link GitChange} instance converted from the given {@link DiffEntry}
+     * @throws IOException JGit library exception
      */
-    private static GitChange convertDiffEntryToGitChange(DiffEntry diffEntry) {
+    private static GitChange convertDiffEntryToGitChange(
+            DiffEntry diffEntry, DiffFormatter formatter) throws IOException {
+        final List<Integer> lines = formatter.toFileHeader(diffEntry).toEditList().stream()
+                .flatMapToInt(edit -> Arrays.stream(
+                        new IntRange(edit.getBeginB(), edit.getEndB() - 1).toArray()))
```
rnveach: So this will convert `1`,`4` to `1,2,3`?

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -203,4 +204,26 @@ public void testParseFilePermissionChange() throws Exception {
...
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            Files.write(helloWorld.toPath(), "line 0\nline 1 change\nline X\n"
+                    .getBytes(Charset.forName("UTF-8")), StandardOpenOption.WRITE);
+            GitUtils.addAllAndCommit(repository, "change line 1, delete line 2, add a line");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be 1 change detected", 1, changes.size());
+            final GitChange expected = ImmutableGitChange.builder()
+                    .path("HelloWorld")
+                    .addLines(1, 2)
```
rnveach: So these are the new/changed lines in the new file?
So we lost out on information like what lines were removed from old file? So complete removal of a line shows no line changes?

Please make 3 new tests to confirm this behavior: adding new line, and removing line, changing existing line.

rnveach:
> Since IntStream.range is banned for the cobertura bug, I use apache.common's IntRange instead.

I added banned classes to import control. Please rebase.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
Luolc: Yes. I mentioned this at the post:
> we have to use for-loop instead of map, for convertDiffEntryToGitChange now throws IOException

There can't be checked exception in lambda

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
Luolc: Yes.

src/main/java/com/github/checkstyle/regression/git/DiffParser.java
rnveach: I didn't know this. We can ignore this for now.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
Luolc: I split this UT to three, including adding line, removing line and changing line.

Luolc:
> I added banned classes to import control. Please rebase.

Rebased.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
```diff
@@ -203,4 +204,69 @@ public void testParseFilePermissionChange() throws Exception {
...
+                    .getBytes(Charset.forName("UTF-8")), StandardOpenOption.APPEND);
+            GitUtils.addAllAndCommit(repository, "add original three lines");
+            GitUtils.createNewBranchAndCheckout(repository, "foo");
+            Files.write(helloWorld.toPath(), "line 0\n"
+                    .getBytes(Charset.forName("UTF-8")), StandardOpenOption.TRUNCATE_EXISTING);
+            GitUtils.addAllAndCommit(repository, "remove two lines");
+            final List<GitChange> changes = DiffParser.parse(
+                    repository.getDirectory().getParent(), "foo");
+            assertEquals("There should be 1 change detected", 1, changes.size());
+            final GitChange expected = ImmutableGitChange.builder()
```
rnveach: As new test shows, removed lines are completely invisible from view.
This can stay for now, but I think it has to change as it will cause issues for some types of regression.

How will we know what method/code/comment was changed if no lines are displayed?
Issues that will be affected:
https://github.com/checkstyle/regression-tool/issues/6 - need to know which method
https://github.com/checkstyle/regression-tool/issues/49 - need to know if has changed besides chmod
https://github.com/checkstyle/regression-tool/issues/69 - need to know if was comment or code

I think we have to work with both sides of the file and not just the latest file.
@Luolc Do you agree? If so, let's create a new issue, unless you want to try and fix it now.

Luolc: I split `lines` to `addedLines` and `deletedLines` now. Is that OK?

src/main/java/com/github/checkstyle/regression/data/GitChange.java
```diff
@@ -32,4 +34,18 @@
      * @return the path of the changed file
      */
     String path();
+
+    /**
+     * The line numbers of the added codes.
```
rnveach: What is `code` here? We can't be sure new lines are code, comments, etc...
Should this just be changes?

src/main/java/com/github/checkstyle/regression/data/GitChange.java
```diff
@@ -32,4 +34,18 @@
...
+     * The first line of a file is marked as line zero.
+     * @return the line numbers of the added codes
+     */
+    List<Integer> addedLines();
+
+    /**
+     * The line numbers of the deleted codes.
```
rnveach: Same here.

src/test/java/com/github/checkstyle/regression/git/DiffParserTest.java
rnveach: I assume this should be ok.
We will probably eventually need some way to examine the contents of these lines and their surrounding text for processing.

rnveach: I assume this should be ok.
Eventually we will probably need some way to examine the before and after file for it's contents and surrounding context.

src/main/java/com/github/checkstyle/regression/data/GitChange.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/data/GitChange.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/data/GitChange.java
```diff
@@ -32,4 +34,18 @@
      * @return the path of the changed file
      */
     String path();
+
+    /**
+     * The line numbers of the added changes.
+     * The first line of a file is marked as line zero.
+     * @return the line numbers of the added codes
```
rnveach: You missed other codes.

src/main/java/com/github/checkstyle/regression/data/GitChange.java
```diff
@@ -32,4 +34,18 @@
...
+     */
+    List<Integer> addedLines();
+
+    /**
+     * The line numbers of the deleted changes.
+     * The first line of a file is marked as line zero.
+     * @return the line numbers of the deleted codes
```
rnveach: same.

src/main/java/com/github/checkstyle/regression/data/GitChange.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/data/GitChange.java
Luolc: Done.

PR #83, opened 2017-08-16
minor: added forbidden classes because of code coverage

rnveach:
Copied from Checkstyle.


rnveach:
@Luolc I am merging this to be sure we don't accidentally add these classes.

Issue #78, opened 2017-08-16
extract: grab property info


Luolc:
We need to update our extract info injection code. To generate the property info as well.

reference in checkstyle: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/internal/XdocsPagesTest.java#L931

rnveach:
Please expand issue to what information we will be obtaining.

Luolc:
Updated the first post.

PR #84, opened 2017-08-21
Issue #78: extract: grab property info

Luolc:
#78 

Add property name and type. Example is uploading.


rnveach:
> Other enums are all seen as string. I haven't find a proper way to handle all the enums currently.

This is a limitation of Checkstyle.
See: 
https://github.com/checkstyle/checkstyle/issues/3542
https://github.com/checkstyle/checkstyle/issues/3543
https://github.com/checkstyle/checkstyle/issues/3254

> Now type could be detected as follows:
> String

Everything is recognized as String if it isn't identified as one of the others. This doesn't mean it is truly String.

rnveach:
@Luolc Until the issues I mentioned above are done, I don't think we can rely fully on the type we get from Checkstyle. Even if we do get all types, we would need more reflection to get enumeration values as they could change between releases or in PR.

Did you plan to use the types gathered and how?

config/sevntu_suppressions.xml
```diff
@@ -4,5 +4,5 @@
         "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-    <suppress checks="MultipleStringLiteralsExtended" files=".*[\\/]src[\\/]test[\\/]"/>
+    <suppress checks="MultipleStringLiteralsExtended" files="[\\/]ExtractInfoGeneratorTest.java$|.*[\\/]src[\\/]test[\\/]"/>
```
rnveach: If we are not going to fix this for any custom tests, put path and Test name as suppression.
Example: `.*extract[\\/]\w+Test\.java`

config/suppressions.xml
```diff
@@ -13,7 +13,7 @@
     <suppress checks="MagicNumber" files=".*[\\/]src[\\/]test[\\/]"/>
     <suppress checks="AvoidStaticImport" files=".*[\\/]src[\\/]test[\\/]"/>
     <suppress checks="WriteTag" files=".*[\\/]src[\\/]test[\\/]"/>
-    <suppress checks="MultipleStringLiterals" files=".*[\\/]src[\\/]test[\\/]"/>
+    <suppress checks="MultipleStringLiterals" files="[\\/]ExtractInfoGeneratorTest.java$|.*[\\/]src[\\/]test[\\/]"/>
```
rnveach: same.

config/findbugs-exclude.xml
```diff
@@ -27,4 +27,9 @@
         <Class name="~.*\.Immutable.*"/>
         <Bug pattern="NP_METHOD_PARAMETER_TIGHTENS_ANNOTATION"/>
     </Match>
+    <Match>
+        <!-- We can't change generated source code. -->
+        <Class name="~.*\.GsonAdapters.*"/>
```
rnveach: Where is this file/class? I can't remember.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -40,12 +47,57 @@
...
+
+    /** Properties of abstract Javadoc check. */
+    private static final Set<String> JAVADOC_CHECK_PROPERTIES =
+            getProperties(AbstractJavadocCheck.class);
+
+    /** Properties of abstract file-set check. */
+    private static final Set<String> FILESET_PROPERTIES = getProperties(AbstractFileSetCheck.class);
+
+    /** Properties without document. */
+    private static final List<String> UNDOCUMENTED_PROPERTIES = Arrays.asList(
```
rnveach: We are not documenting properties here.
This is more a list of fields that aren't properties.

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -40,12 +47,57 @@
  * @author LuoLiangchen
  */
 public class ExtractInfoGeneratorTest {
+    /** Modules which do not have global properties to drop. */
+    private static final List<String> XML_FILESET_LIST = Arrays.asList(
```
rnveach: Can we drop this and use the check's class hierarchy to determine the parent?

src/main/resources/com/github/checkstyle/regression/extract/ExtractInfoGeneratorTest.java
```diff
@@ -94,6 +148,116 @@ else if (ModuleReflectionUtils.isRootModule(clazz)) {
         object.add("interfaces", interfaces);
         object.add("hierarchies", hierarchies);
 
+        final JsonUtil.JsonArray properties = new JsonUtil.JsonArray();
+        for (String propertyName : getNecessaryProperties(clazz)) {
```
rnveach: Add empty line before `for` and after `}` for `for`.

rnveach:
@Luolc ping

rnveach:
@Luolc ping

Issue #79, opened 2017-08-16
Processing UT changes to get possible property values


Luolc:
We could process the change of UT, find the changed method, and collect the property value from the config attribute.

Example PRs:
https://github.com/checkstyle/checkstyle/pull/4905
https://github.com/checkstyle/checkstyle/pull/4902
https://github.com/checkstyle/checkstyle/pull/4853
https://github.com/checkstyle/checkstyle/pull/4770
https://github.com/checkstyle/checkstyle/pull/4736

> We can assume the tests will be in a specific location as it should just be a change of `main` to `test` and appending `Test` to the class name.

rnveach:
> process the change of UT, find the changed method

Like I said in google groups, We can't always expect there to be test changes with production code. If it is just a fixing of code and not a logic change or such, we still need to run regression on it and all it's properties to verify nothing changed or is broken.
Also why do we need to only grab the property values of changed UTs. If their values didn't change in regression, then it will agree with no changes in the UTs. If they show differences, than that is unexpected since they didn't make changes to those UTs and they should add them in their PR then. We should grab all values from UTs, and not just ones that changed.

I feel we will get better results if we grab all possible values and not ones that changed.
The only properties that probably may not produce any results will be regular expressions for file/folder paths.

Luolc:
> Also why do we need to only grab the property values of changed UTs

I am OK with this. Whatever only changed UTs or all UTs, we would implement the same function that grab the property info. And I think grabbing from all UTs is good.

PR #85, opened 2017-08-21
Issue #79: Processing UT changes to get possible property values

Luolc:
#79 

This PR includes a custom check and the utility class to run the check.

The check walk through all `@Test` annotation to target the UT method, find the statement like `final DefaultConfiguration fooModuleConfig = createModuleConfig(FooCheck.class)`, and get properties by detecting the `fooModuleConfig.addAttribute` method call.

If this is OK, then we could integrate this into our generator.


src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
```diff
@@ -0,0 +1,211 @@
...
+
+import com.github.checkstyle.regression.data.ImmutableProperty;
+import com.github.checkstyle.regression.data.ModuleInfo;
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * The custom check which processes the unit test class of a checkstyle module.
```
rnveach: Add extra blank line between summary and details.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
```diff
@@ -0,0 +1,211 @@
...
+ * This check would walk through the {@code @Test} annotation, find variable definition
+ * like {@code final DefaultConfiguration checkConfig = createModuleConfig(FooCheck.class)}
+ * and grab the property info from {@link DefaultConfiguration#addAttribute(String, String)}
+ * method call.
```
rnveach: So we won't support tests where `checkConfig` is defined as a field, instead of a local variable?
It is not  possible to support both?

If so, we need changes in Checkstyle as there are loads of different ways tests are made and they must be organized.
Please make a new issue in checkstyle and let's get @romani 's approval.

If we are going this way, please update Javadoc with example of how test classes should be defined.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+            throws CheckstyleException, ReflectiveOperationException {
+        final DefaultConfiguration moduleConfig = createModuleConfig(UnitTestProcessorCheck.class);
+        final Configuration dc = createTreeWalkerConfig(moduleConfig);
+        final Checker checker = new Checker();
+        checker.setModuleClassLoader(Thread.currentThread().getContextClassLoader());
+        checker.configure(dc);
+        final List<File> processedFiles = Collections.singletonList(new File(path));
+        checker.process(processedFiles);
+
+        final UnitTestProcessorCheck check = getCheckInstance(checker);
```
rnveach: Truthfully we don't need Check's instance. We just need `unitTestToPropertiesMap`.
Why don't we make field static and create a getter for it? Than we can just do `UnitTestProcessorCheck.getUnitTestToPropertiesMap` instead of doing reflection hacks.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
```diff
@@ -0,0 +1,211 @@
...
+            if (configVariableName.isPresent()) {
+                final Set<ModuleInfo.Property> properties = new LinkedHashSet<>();
+
+                for (DetailAST expr : getAllChildrenWithToken(methodBlock, TokenTypes.EXPR)) {
+                    if (isAddAttributeMethodCall(expr.getFirstChild(), configVariableName.get())) {
+                        final DetailAST elist =
+                                expr.getFirstChild().findFirstToken(TokenTypes.ELIST);
+                        final String key = convertExprToText(elist.getFirstChild());
+                        final String value = convertExprToText(elist.getLastChild());
+                        properties.add(ImmutableProperty.builder().name(key).value(value).build());
```
rnveach: What if value is non-string?
Will this cause problems with the property value? We have some examples in checkstyle that use enumeration and do something like `toString()`.

src/test/resources/com/github/checkstyle/regression/module/ReturnCountCheckTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtils;
+
+public class ReturnCountCheckTest extends AbstractModuleTestSupport {
```
rnveach: This confused me at first as I was expecting a test file for `UnitTestProcessorCheck` and was wondering where `ReturnCount` came from.
Does this need to be `ReturnCount`? Can we turn it into some dummy name like `InputModuleExampleTest`?

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Processes the unit test class with the given path.
+     * @param path the path of the unit test class
+     * @return the unit test method name to properties map
+     * @throws CheckstyleException          failure when running the check
+     * @throws ReflectiveOperationException failure of reflection on checker and tree walker
+     */
+    public static Map<String, Set<ModuleInfo.Property>> process(String path)
```
rnveach: This class has 2 functions, executing a custom check and getting the results for the specific check.
I assume we may need to write another custom check for https://github.com/checkstyle/regression-tool/issues/6 .

If you agree, I think we should rename this class to something more generic, have it take the module class as input, and return nothing as everything will be retrieved through static getters.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.module;
```
rnveach: This should move with the custom check too.

config/pmd.xml
```diff
@@ -34,7 +34,11 @@
     <!-- we use class comments as source for xdoc files, so content is big and that is by design -->
     <exclude name="CommentSize"/>
   </rule>
-  <rule ref="rulesets/java/comments.xml/CommentRequired" />
+  <rule ref="rulesets/java/comments.xml/CommentRequired">
+    <properties>
+      <property name="publicMethodCommentRequirement" value="Ignored"/>
```
rnveach: Please explain this change.

pom.xml
```diff
@@ -394,6 +394,11 @@
                                         <lineRate>0</lineRate>
                                     </regex>
                                     <regex>
+                                        <pattern>com.github.checkstyle.regression.module.UnitTestProcessorCheck</pattern>
+                                        <branchRate>76</branchRate>
+                                        <lineRate>92</lineRate>
```
rnveach: While you are making changes, please post this file's code coverage report.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
Luolc: > So we won't support tests where checkConfig is defined as a field, instead of a local variable?
It is not possible to support both?

I think it could be possible. I didn't realize the config could be a field before. I would have a try on this.

rnveach: @Luolc Please use your check and scan Checkstyle's repo. Any true module tests where your check finds nothing or not enough is an area you need to look into.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
Luolc: It would have problems with non-string.

I viewed some test just now. I see smth like https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheckTest.java#L65.

Do you remember anything else that is non-string or `toString()`?

I would scan checkstyle repo and pay attention to this as well.

rnveach: I don't remember anything else specifically but anything you can code you can find.

If this is becoming too much to support, then in our issue to checkstyle, let's ask that they conform to a specific style.
We would need to create a list of required styles, make changes to them, and maybe eventually make a custom check that would enforce this in their repository.

rnveach: > WrapOption.EOL.toString()

Easy way to support this is to find anything like `IDENT1.IDENT2.toString()` and save value as `IDENT2`, unless you see some downside or false positive.

rnveach: @Luolc Something else we have to watch out for is exception cases.
Most enumerations will have atleast 1 exception case where the value specified is not an enumerated value and is there just to test the exception message.

rnveach:
@Luolc Remember to ping me when you have made fixes and reply as `Done` to each point.
We don't get updates when new commits have been pushed.

config/pmd.xml
```diff
@@ -34,7 +34,11 @@
     <!-- we use class comments as source for xdoc files, so content is big and that is by design -->
     <exclude name="CommentSize"/>
   </rule>
-  <rule ref="rulesets/java/comments.xml/CommentRequired" />
+  <rule ref="rulesets/java/comments.xml/CommentRequired">
+    <properties>
+      <property name="violationSuppressXPath" value="//Annotation/MarkerAnnotation//Name[@Image='Override']"/>
```
rnveach: Please explain this change.

pom.xml
rnveach: Still not done.

src/main/java/com/github/checkstyle/regression/customcheck/CustomCheckProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+ * the collected property info.</p>
+ * @author LuoLiangchen
+ */
+public final class CustomCheckProcessor {
+    /** Prevents instantiation. */
+    private CustomCheckProcessor() {
+    }
+
+    /**
+     * Processes the file with the given path using the given custom check.
```
rnveach: Update documentation that check needs a public/static way to retrieve it's own results using this method.

src/main/java/com/github/checkstyle/regression/customcheck/UnitTestProcessorCheck.java
```diff
@@ -0,0 +1,220 @@
...
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * The custom check which processes the unit test class of a checkstyle module and grab the
+ * possible properties that could be used for generating config.
+ *
+ * <p>This check would walk through the {@code @Test} annotation, find variable definition
+ * like {@code final DefaultConfiguration checkConfig = createModuleConfig(FooCheck.class)}
+ * and grab the property info from {@link DefaultConfiguration#addAttribute(String, String)}
+ * method call.</p>
```
rnveach: https://github.com/checkstyle/regression-tool/pull/85#discussion_r134472709
> If we are going this way, please update Javadoc with example of how test classes should be defined.

Not done.

src/main/java/com/github/checkstyle/regression/customcheck/UnitTestProcessorCheck.java
```diff
@@ -0,0 +1,220 @@
...
+        };
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        if ("Test".equals(ast.findFirstToken(TokenTypes.IDENT).getText())) {
+            final DetailAST methodDef = ast.getParent().getParent();
+            final DetailAST methodBlock = methodDef.findFirstToken(TokenTypes.SLIST);
+            final Optional<String> configVariableName =
+                    getModuleConfigVariableName(methodBlock);
```
rnveach: https://github.com/checkstyle/regression-tool/pull/85#discussion_r134472709
> So we won't support tests where checkConfig is defined as a field, instead of a local variable?
It is not possible to support both?

I don't see an update or any tests for this.

Luolc:
@rnveach 
I am still working on https://github.com/checkstyle/regression-tool/pull/85#discussion_r134472709 and was thinking to ping you when finish this together. And I found I might spend more time so I first commit part of the work.

I scan and output all the property info from UTs, and I have to check them manually whether the processor missed anything or produce anything wrong.

I am working on the config as class field and the enum property value, but I am still not very confident that they are absolutely good now.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
```diff
@@ -0,0 +1,211 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.checkstyle.regression.module;
```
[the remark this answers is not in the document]
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessorCheck.java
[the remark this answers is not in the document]
Luolc: Done.

Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessor.java
Luolc: Done.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessor.java
Luolc: Updated.

src/main/java/com/github/checkstyle/regression/module/UnitTestProcessor.java
Luolc: Changed to static method now.

src/test/resources/com/github/checkstyle/regression/module/ReturnCountCheckTest.java
Luolc: Done. This class is a copy of `ReturnCountCheckTest` from checkstyle repo. I didn't change its name at first.

config/pmd.xml
Luolc: Without this pmd would force us to add javadoc on `@Override` method. I copied this settings from checkstyle repo.

rnveach:
> Checking the result of all checkstyle module UTs may still take a lot of time I think.

I would try automating it somehow and not worry about manually checking every single digit.

For example, I said if you run your check does any file not produce any results?
We have reflection code that lists all module properties. Use that to against the results of your check to see if any properties were missed or if any extra were picked up.
Put breaks in your check to see what type of property values are not being selected, but match the other criteria.

I already gave you some examples of other areas that your check doesn't cover, so we can go with that for now. Those should be the most popular ways to code the tests.

config/pmd.xml
rnveach: Ok, thanks.
This can be ignored.

rnveach:
@Luolc ping

Luolc:
@rnveach 
There are still many modules left now.
I wrote this for helping me analyze the UTs: https://github.com/Luolc/regression-tool/blob/ut-scan/src/test/java/com/github/checkstyle/regression/customcheck/UnitTestProcessorCheckTest.java#L101. I checked by the assertion and break points in the code.

I made some updates and could also support value as:
- Strings joining with `+`, like `"CLASS_DEF" + ",VARIABLE_DEF"`
- Enums like `BlockOption.TEXT.toString()`. Here I used regex to handle this. So it might have problems if the expression is in multiple lines.
- module config declared as class field, and **assigned in the single UT method**. I would explain this at below.

The problems found currently:
1.`BeforeExecutionExclusionFileFilter`: the UT class name is not `BeforeExecutionExclusionFileFilterTest` but `ExclusionBeforeExecutionFileFilterTest`. 

2.The property value could be a variable: 
current tests found: 
- AbstractClassNameCheck
- AtclauseOrderCheck
- CatchParameterNameCheck

example: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/naming/AbstractClassNameCheckTest.java#L66

3.The property value could be a variable: 
current tests found: 
- HeaderCheck

example: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/header/HeaderCheckTest.java#L63

4.Some UTs missed tests of some properties
current tests found: 
- ClassFanOutComplexityCheck
- ConstantNameCheck

example: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/naming/ConstantNameCheckTest.java , no test for property `applyToPrivate`

5.different way of class field module config
I mentioned above that we could support a kind of class field config now.
example: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/coding/DefaultComesLastCheckTest.java#L43
The variable could be declared as class field, **but we must re-assign it every time in each UT method**

There are many other UTs with class field config that is assign in `setUp`. I haven't done with this yet.

6.Some UTs don't use `createModuleConfig`
- FileTabCharacterCheck: createConfig
- BeforeExecutionExclusionFileFilter: createBeforeExecutionFileFilterConfig

we may need to collect names of all these kind of method.

\-\-\-\-\-\-
My checking is at `HeaderCheck` now (in a Lexicographic order). 40 modules. While we have 160+ modules, there are still too many. And I have already found a lot of different kind of situation that cannot support currently. I am afraid I still need much time on this if we want to solve them.
This is the list of them:
```
com.puppycrawl.tools.checkstyle.checks.naming.AbbreviationAsWordInNameCheck
com.puppycrawl.tools.checkstyle.checks.naming.AbstractClassNameCheck
com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationLocationCheck
com.puppycrawl.tools.checkstyle.checks.annotation.AnnotationUseStyleCheck
com.puppycrawl.tools.checkstyle.checks.sizes.AnonInnerLengthCheck
com.puppycrawl.tools.checkstyle.checks.coding.ArrayTrailingCommaCheck
com.puppycrawl.tools.checkstyle.checks.ArrayTypeStyleCheck
com.puppycrawl.tools.checkstyle.checks.javadoc.AtclauseOrderCheck
com.puppycrawl.tools.checkstyle.checks.AvoidEscapedUnicodeCharactersCheck
com.puppycrawl.tools.checkstyle.checks.coding.AvoidInlineConditionalsCheck
com.puppycrawl.tools.checkstyle.checks.blocks.AvoidNestedBlocksCheck
com.puppycrawl.tools.checkstyle.checks.imports.AvoidStarImportCheck
com.puppycrawl.tools.checkstyle.checks.imports.AvoidStaticImportCheck
com.puppycrawl.tools.checkstyle.checks.metrics.BooleanExpressionComplexityCheck
com.puppycrawl.tools.checkstyle.checks.naming.CatchParameterNameCheck
com.puppycrawl.tools.checkstyle.checks.metrics.ClassDataAbstractionCouplingCheck
com.puppycrawl.tools.checkstyle.checks.naming.ClassTypeParameterNameCheck
com.puppycrawl.tools.checkstyle.checks.indentation.CommentsIndentationCheck
com.puppycrawl.tools.checkstyle.checks.coding.CovariantEqualsCheck
com.puppycrawl.tools.checkstyle.checks.imports.CustomImportOrderCheck
com.puppycrawl.tools.checkstyle.checks.metrics.CyclomaticComplexityCheck
com.puppycrawl.tools.checkstyle.checks.coding.DeclarationOrderCheck
com.puppycrawl.tools.checkstyle.checks.coding.DefaultComesLastCheck
com.puppycrawl.tools.checkstyle.checks.DescendantTokenCheck
com.puppycrawl.tools.checkstyle.checks.design.DesignForExtensionCheck
com.puppycrawl.tools.checkstyle.checks.blocks.EmptyBlockCheck
com.puppycrawl.tools.checkstyle.checks.blocks.EmptyCatchBlockCheck
com.puppycrawl.tools.checkstyle.checks.whitespace.EmptyLineSeparatorCheck
com.puppycrawl.tools.checkstyle.checks.coding.EmptyStatementCheck
com.puppycrawl.tools.checkstyle.checks.coding.EqualsAvoidNullCheck
com.puppycrawl.tools.checkstyle.checks.coding.EqualsHashCodeCheck
com.puppycrawl.tools.checkstyle.checks.sizes.ExecutableStatementCountCheck
com.puppycrawl.tools.checkstyle.checks.coding.ExplicitInitializationCheck
com.puppycrawl.tools.checkstyle.checks.coding.FallThroughCheck
com.puppycrawl.tools.checkstyle.checks.sizes.FileLengthCheck
com.puppycrawl.tools.checkstyle.checks.design.FinalClassCheck
com.puppycrawl.tools.checkstyle.checks.coding.FinalLocalVariableCheck
com.puppycrawl.tools.checkstyle.checks.FinalParametersCheck
com.puppycrawl.tools.checkstyle.checks.whitespace.GenericWhitespaceCheck
com.puppycrawl.tools.checkstyle.checks.header.HeaderCheck
```

rnveach:
> Enums like BlockOption.TEXT.toString(). Here I used regex to handle this. So it might have problems if the expression is in multiple lines.

Why regular expression? If it is in the form `XXX.YYY.toString()` can't we assume what we want is the IDENT `YYY`?

> BeforeExecutionExclusionFileFilter: the UT class name is not BeforeExecutionExclusionFileFilterTest but ExclusionBeforeExecutionFileFilterTest.

I see `filefilters.BeforeExecutionExclusionFileFilter` in production, and `filters.BeforeExecutionFileFilterSetTest` and `filefilters.ExclusionBeforeExecutionFileFilterTest` in test.
Let's make an issue of this in Checkstyle. We need tests named after their classes they are testing. Something should be named `BeforeExecutionExclusionFileFilterTest`.

> 2. The property value could be a variable

Let's split this off into another issue. The example you gave is pretty simple and the variable could be inlined. The check could also be modified to just track variable assignments and pickup the very basic uses.

> 3.The property value could be a variable

Move this to another issue too. I think we should recognize `getPath("String")` as we need something to run regression on when the input is a file.

> Some UTs missed tests of some properties

This must be an issue in Checkstyle.
There is nothing we can do if they miss UT on their own properties.
We will still pick up property in https://github.com/checkstyle/regression-tool/pull/84 , but picking a value to use for it might be harder. If we have a property to use in the configuration, but can't decide on a value, we should print some type of warning in the console of our application and drop the property.
For CIs we can make an additional option to fail the build if this happens.

> 6.Some UTs don't use createModuleConfig

I still think we should create an issue in Checkstyle to have them organize their UTs more.

> FileTabCharacterCheck: createConfig
> BeforeExecutionExclusionFileFilter: createBeforeExecutionFileFilterConfig

Especially these 2. I don't see a reason right now why they can't be standardized.

> we may need to collect names of all these kind of method.

Yes, we will need to do this for any issues we create. It would be better if we could atleast create an automated way to detect these instances and not rely on manually reading each one.

rnveach:
@Luolc 
> It would be better if we could atleast create an automated way to detect these instances and not rely on manually reading each one.

Is it possible to automate finding of bad cases, even if it only finds like 80% of the cases?
For example:
if a UT method (with annotation `@Test`) has `verify` method call, then it must have variable definition `DefaultConfiguration XXX` and if it does `XXX.addAttribute` then the parameters must be a String, `getPath("String")`, or an enum in the form we have discussed above.

Luolc:
> Why regular expression? If it is in the form XXX.YYY.toString() can't we assume what we want is the IDENT YYY?

I don't know if we have smth like `XXX.YYY.ZZZ.toString()`, or full qualified name ones. Regex could handle them instead. If we finally make sure that we only have one format, then getting the value from ast could be better.

> Let's split this off into another issue. The example you gave is pretty simple and the variable could be inlined. The check could also be modified to just track variable assignments and pickup the very basic uses.

OK. That makes sense.

> Move this to another issue too. I think we should recognize getPath("String") as we need something to run regression on when the input is a file.

Although `getPath` is a method of super class. It is possible to be override anyway. 

> Is it possible to automate finding of bad cases, even if it only finds like 80% of the cases?

Maybe I need another custom check, or at least some kind of regex way for this.
My current way is still low-efficient.

I would sum up the issues we need to create both here and in checkstyle repo and create them together sooner.

rnveach:
> Maybe I need another custom check

yes, they would need to be in sevntu as it is just for us.

> Although getPath is a method of super class. It is possible to be override anyway.

Yes, to get the correct path would would have to inspect the `getPackageLocation` method which should always be a simple `return "string"` now that we are rewriting tests some.

> I don't know if we have smth like XXX.YYY.ZZZ.toString()

I doubt it, but something like the custom check I mentioned in previous post to validate UTs that use `verify` would find out if there was one like this pretty fast as it will print a violation on it.

I think it makes more and more sense we go this route and force checkstyle use the variants we approve. This will make things easier on our custom check for all the weird cases we could possibly see.

rnveach:
@Luolc ping

rnveach:
> 1.BeforeExecutionExclusionFileFilter: the UT class name is not BeforeExecutionExclusionFileFilterTest but ExclusionBeforeExecutionFileFilterTest.

Issue #5104 in checkstyle was created for this.

rnveach:
Point 1 is nearly fixed in Checkstyle https://github.com/checkstyle/checkstyle/issues/5104 .
Point 2 and 5 is now an issue in Sevntu https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/610.
Point 3 should just be a case we need to support. All path locations have been standardized in Checkstyle, so we should be able to find the input pretty easily.
Point 4 is now an issue in Checkstyle https://github.com/checkstyle/checkstyle/issues/5156 .
Point 6 is now an issue in Checkstyle https://github.com/checkstyle/checkstyle/issues/5157.

This PR is on hold until the issues are fixed and uniformity is brought to Checkstyle.
We should be able to use this check to verify issues in Checkstyle are fixed.

rnveach:
Custom check needs to be changed to gather information from multiple configurations in 1 test.
Example:
https://github.com/checkstyle/checkstyle/blob/2aca36156fa461f2d54f1386743a0146bdf8b4bd/src/test/java/com/puppycrawl/tools/checkstyle/checks/header/HeaderCheckTest.java#L236-L247
1 test defines custom HeaderCheck, and defines custom Checker.
We need to distinguish between the Check we want and others. I'm not sure if we can get all values or should.

rnveach:
New check is done and will soon be added to Checkstyle.

Points 1, 2, 5, and 6 are done.
Point 4 can be done at any time.

All that is left is to fix this PR and merge it.

Issue #41, opened 2017-06-27
Update CI from new additions in main Project


rnveach:
Checkstyle has been updating PMD and other inspection tools adding in new checks we didn't have before.
We should consider adding them to regression-tool too.

rnveach:
Besides the open PRs, all that is left is updating PMD configuration.

rnveach:
Only eclipse CI needs to be fixed.

PR #87, opened 2017-11-14
Issue #41: upgraded maven versions

rnveach:
Issue #41

I couldn't upgrade Checkstyle as it was causing some issue with the tests. I need to examine it more.


rnveach:
Rebased.

PR #90, opened 2017-11-14
Issue #41: added eclipse ci

rnveach:
Issue #41

@romani We should update SH file to be able to get eclipse settings from a different project so we don't have to copy and paste it.

### DIFF
--- a/.ci/eclipse-compiler-javac.sh
+++ b/.ci/eclipse-compiler-javac.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+set -e
+
+if [ -z "$1" ]; then
+    echo "No parameters supplied!"
+    echo "      The classpath of the project and it's libraries to compile must be supplied."
+    exit 1
+fi
+
+ECJ_JAR="ecj-4.7.jar"
+ECJ_MAVEN_VERSION="R-4.7-201706120950"
+ECJ_PATH=~/.m2/repository/$ECJ_MAVEN_VERSION/$ECJ_JAR
+
+if [ ! -f $ECJ_PATH ]; then
+    echo "$ECJ_PATH is not found, downloading ..."
+    mkdir -p $(dirname "$ECJ_PATH")
+    wget http://ftp-stud.fht-esslingen.de/pub/Mirrors/eclipse/eclipse/downloads/drops4/$ECJ_MAVEN_VERSION/$ECJ_JAR -O $ECJ_PATH
+fi
+
+mkdir -p target/classes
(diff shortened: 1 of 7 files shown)


rnveach:
> java.io.IOException: Cannot run program "/home/travis/build/checkstyle/regression-tool/.ci/eclipse-compiler-javac.sh" (in directory "/home/travis/build/checkstyle/regression-tool"): error=13, Permission denied

Windows doesn't mark it as executable. I need to switch to linux to fix.

rnveach:
@romani @Nimfadora 
> The type ImmutableGitChange is already defined

Do either of you know how to fix eclipse issue?
Files created by gson are placed into `<base>\com\...` instead of `<base>\target\generated-sources\annotations\...`.

If I remove `target/generated-sources/annotations` from the SH so eclipse doesn't pull it in, then I get errors on that auto-generated code that should be ignored. I tried adding it to the `nowarn` with no change.
> 4. ERROR in /media/checkstyle/work/regression-tool/com/github/checkstyle/regression/data/ImmutableGitChange.java (at line 237)

romani:
>update SH file to be able to get eclipse settings from a different project so we don't have to copy and paste it.

Nothing is bad in such copy paste. It is better to keep the separate.

rnveach:
@romani @Nimfadora ping.
I still need help with https://github.com/checkstyle/regression-tool/pull/90#issuecomment-344386287

romani:
@rnveach , please do copy paste of script and proceed, too much activity in projects to invest time in this issue

rnveach:
@romani 
> please do copy paste of script and proceed

This is the copy paste of the script and it is failing for the reasons I stated above.

rnveach:
@romani ping
I can try upgrading the copy/paste but I don't have any reason to believe it will help.

romani:
> We should update SH file to be able to get eclipse settings from a different project so we don't have to copy and paste it.

Ok, let's update sh file in main project to take config from some env variable, if variable is not defined use default file ( that is now a config in main project)

rnveach:
@romani It is not a rule failure, it is a compile failure.

> 1. ERROR in /home/travis/build/checkstyle/regression-tool/com/github/checkstyle/regression/data/ImmutableGitChange.java (at line 28)
	public final class ImmutableGitChange implements GitChange {
	                   ^^^^^^^^^^^^^^^^^^
The type ImmutableGitChange is already defined

romani:
judging status of project, lets skpi this CI execution for now, we will start investing in code quality ones project become usable.

Issue #58, opened 2017-07-23
Codecov: Main


Luolc:
We need to add test to appease the codecov for class `Main`

PR #92, opened 2017-11-15
Issue #58: added some code coverage for Main

rnveach:
Issue #58

`Main`'s `main` method was changed to `String...` to make it easier to pass arguments.
Exception exit was changed to `-1` to mimic checkstyle when error occurs.

Code coverage is raised to 75% for this class.


rnveach:
Same issue as other PR. If problem continues, I'll try to find a new commit.

PR #93, opened 2017-11-15
 Issue #41: updated pmd validation 

rnveach:
Issue #41


rnveach:
I am going to merge this but travis failure looks like regression was set to too big a change and is causing out of memory, but I don't know why the PR changing it didn't fail then.
From logs, it looks like it still did regression on old commits as list of modules don't match the commits but I don't know why.
https://travis-ci.org/checkstyle/regression-tool/jobs/302267234#L1024
https://travis-ci.org/checkstyle/regression-tool/jobs/302097214 (must download logs to see)

PR #95, opened 2018-05-26
upgrade dependencies

rnveach:


romani:
it might be good to set up CRON execution - ones a week
and make badge for README to easily catch red state

rnveach:
No one has been working on this project. State was good before, but failure started with memory issue that couldn't be fixed, atleast not yet. See https://github.com/checkstyle/regression-tool/pull/94 .
Other tests verify tool works with latest checkstyle version which is only run when PR is run.
Cron won't help us until we finish this tool and start really using it.

PR #106, opened 2021-01-19
minor: fix travis build

nmancus1:
Fixes failure on travis build; diff.groovy requires config to be in the same directory for crosslinked file generation.  Example of failure: https://travis-ci.org/github/checkstyle/regression-tool/jobs/755185633

image


nmancus1:
Cannot workaround limitation of https://github.com/checkstyle/contribution/issues/539, that issue will fix this one.

